PR #28, opened 2023-07-04
[WIP] Allow to ignore specific errors for specific files

kolyshkin:
This allows to exclude specific known errors for specific files, meaning
that instead of skipping all checks for a particular file, we can only
skip some specific ones (or, rather, ignore their results).
    
The rules can be specified globally, or per-tag, or per-component,
or per-rpm.
    
TODO: convert configs to use this.


kolyshkin:
> getting
> 
> ```
> ./errors.go:33:16: undefined: KnownErrors
> ./main.go:93:22: undefined: KnownErrors
> ```

For now, you can run `go generate` (since I can't update this PR at the moment).

PR #39, opened 2023-07-10
add warning support and --fail-on-warnings flag

rphillips:
Changes strictfipsruntime flag to be a warning and adds warning error support to the application.

Additionally adds `--fail-on-warnings` flag to preserve warnings as errors.

### DIFF
--- a/main.go
+++ b/main.go
@@ -50,6 +50,7 @@ var (
 	components                            []string
 	configFile, configForVersion          string
 	cpuProfile                            string
+	failOnWarnings                        bool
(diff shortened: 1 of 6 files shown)


kolyshkin (commented):
Maybe we can just add `Warning` bool to `ScanResult` structure? As any bool, it is `false` by default, and we can downgrade an error to a warning by setting it to `true`.

rphillips:
I started down that path, but think we may want different levels later.

rphillips:
We can rethink this when you get back... Going to merge to make these warnings.

PR #62, opened 2023-07-14
chore: CRW-4582 - add exclusions for Dev Spaces statically linked binaries

nickboldt:


nickboldt:
Dropped in favour of https://github.com/openshift/check-payload/pull/63/files

PR #70, opened 2023-07-18
Minor optimizations

kolyshkin:
A few optimizations mostly aimed to reduce LOC (but also reducing the amount of garbage to collect).


kolyshkin:
Rebased
/unhold

PR #73, opened 2023-07-18
feat: do not depend on go binary

kolyshkin:
We happen to already use `buildinfo.ReadFile` which provides all the
information we need (same as obtained from `go version -m`).

Let's use it, avoiding go binary dependency and making things a tad
faster.


mrunalp:
Looks good. We can merge if no regressions on payloads.

PR #74, opened 2023-07-19
chore: CRW-4582 add more exceptions for known issues

nickboldt:
chore: [CRW-4582](https://issues.redhat.com//browse/CRW-4582) add more exceptions for known issues

Signed-off-by: Nick Boldt <nboldt@redhat.com>

### DIFF
--- a/dist/releases/4.10/config.toml
+++ b/dist/releases/4.10/config.toml
@@ -59,6 +59,6 @@ filter_files = [
   "/usr/lib/golang/pkg/tool/linux_amd64/compile",
   "/usr/lib/golang/pkg/tool/linux_ppc64le/compile",
   "/usr/lib/golang/pkg/tool/linux_s390x/compile",
-  ### helm is not FIPS compliant
+  ### helm is not FIPS compliant; exception needed until https://issues.redhat.com/browse/HELM-506 is resolved
   "/usr/bin/helm"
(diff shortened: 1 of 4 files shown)


dist/releases/4.10/config.toml
```diff
@@ -52,3 +52,13 @@ filter_files = [
...
 ]
+
+[payload.devspaces-udi-rhel8-container]
+filter_files = [
+  ### include golang compiler intentionally; exception needed until https://issues.redhat.com/browse/GT-130 is resolved
+  "/usr/lib/golang/pkg/tool/linux_amd64/compile",
+  "/usr/lib/golang/pkg/tool/linux_ppc64le/compile",
+  "/usr/lib/golang/pkg/tool/linux_s390x/compile",
+  ### helm is not FIPS compliant
+  "/usr/bin/helm"
```
rphillips, line 63: We should be able to rebuilt helm with the correct setting?

nickboldt: sure but who is "we" and when will that happen?

rphillips: Can we remove it from the image?

nickboldt: technically, yes. But afaik customers are using it.

PR #75, opened 2023-07-19
rpm-related refactoring and cleanups

kolyshkin:
This is done in preparation to support per-rpm exclusions in payload scan.
Please see individual commit messages for details.

The only notable changes are:
 - `node.` is now `rpm.` in configs;
 - only the rpm name is used in configs.

### DIFF
--- a/internal/scan/node_scan.go
+++ b/internal/scan/node_scan.go
@@ -33,15 +33,15 @@ func RunNodeScan(ctx context.Context, cfg *types.Config) []*types.ScanResults {
 	root := cfg.NodeScan
 	rpms, _ := rpm.GetAllRPMs(ctx, root)
 	for _, pkg := range rpms {
-		tag := NewTag(pkg)
-		files, err := rpm.GetFilesFromRPM(ctx, root, pkg)
+		tag := NewTag(pkg.Name)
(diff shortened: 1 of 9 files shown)


internal/scan/node_scan.go
```diff
@@ -33,15 +33,15 @@ func RunNodeScan(ctx context.Context, cfg *types.Config) []*types.ScanResults {
 	root := cfg.NodeScan
 	rpms, _ := rpm.GetAllRPMs(ctx, root)
 	for _, pkg := range rpms {
-		tag := NewTag(pkg)
-		files, err := rpm.GetFilesFromRPM(ctx, root, pkg)
+		tag := NewTag(pkg.Name)
+		files, err := rpm.GetFilesFromRPM(ctx, root, pkg.NVRA)
```
kolyshkin, line 37: We use name (e.g. `glibc`) for config exceptions, and NVRA (e.g. `glibc-2.28-189.5.el8_6.x86_64`) for getting the list of files (because more than 1 glibc may be installed).

kolyshkin: Potentially, we can modify the code to be able to use both the name and the NVRA (name-version-release.architecture) in the exclusion lists; until there's a need for that, I'd like to not do it.

PR #77, opened 2023-07-20
chore: CRW-4582 exclude /usr/libexec/catatonit/catatonit in payload.devspaces-udi-rhel8-container as it's already excluded in root config.toml

nickboldt:
chore: [CRW-4582](https://issues.redhat.com//browse/CRW-4582) exclude /usr/libexec/catatonit/catatonit in payload.devspaces-udi-rhel8-container as it's already excluded in root config.toml

Signed-off-by: Nick Boldt <nboldt@redhat.com>

### DIFF
--- a/dist/releases/4.10/config.toml
+++ b/dist/releases/4.10/config.toml
@@ -55,6 +55,8 @@ filter_files = [
 
 [payload.devspaces-udi-rhel8-container]
 filter_files = [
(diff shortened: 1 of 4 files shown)


nickboldt:
apply this PR, then
```
make
./check-payload scan image --spec quay.io/devspaces/udi-rhel8:3.7-18   -V 4.12

...
---- Successful run with warnings
```

PR #79, opened 2023-07-21
hotfix: fix 4.9 config

kolyshkin:
Commit 33b5277ec32d resulted in two rpm.cri-o sections.

Remove the duplicate (which results in a panic -- we need to add a CI job to prevent this).


mrunalp (approved):
/lgtm

PR #82, opened 2023-07-21
perf: skip all files with no x bit set

kolyshkin:
Such files are most probably not binaries, or they are binaries that
can't be executed (until `x` bit is set, that is).
    
This helps to skip calling `isElfExe` for most libraries.

Shows a noticeable improvement on my machine.


mrunalp (approved):
/lgtm

PR #83, opened 2023-07-21
Node scan hotfixes

kolyshkin:
This fixes three issues with node scan, two of which prevented node scans from working.

The first one was introduced by #75 and broke all node scans.
The second one was there from the beginning and broke some node scans.
The last issue was not reporting errors from `rpm -qa`.

See individual commits for more details.

Tested to fix the issue of not being able to use node scan.

### DIFF
--- a/internal/rpm/rpm.go
+++ b/internal/rpm/rpm.go
@@ -20,16 +20,19 @@ type Info struct {
 
 func GetFilesFromRPM(ctx context.Context, root, rpm string) ([]string, error) {
 	klog.Infof("rpm -ql %v", rpm)
-	files := []string{}
-	var stdout bytes.Buffer
-	var stderr bytes.Buffer
-	cmd := exec.CommandContext(ctx, "rpm", "-ql", "--root", root, rpm)
+	dbpath, err := rpmDBPath(root)
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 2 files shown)


internal/rpm/rpm.go
```diff
@@ -64,6 +64,12 @@ func GetAllRPMs(ctx context.Context, root string) ([]Info, error) {
 		}
 		rpms = append(rpms, Info{Name: f[0], NVRA: f[1]})
 	}
+	if err := scanner.Err(); err != nil {
+		return nil, fmt.Errorf("error reading rpm -qa: %w", err)
+	}
+	if len(rpms) == 0 {
```
mrunalp, line 70: I am thinking that if this scanner is to work for any image in the future then it should pass on a FROM scratch base image that has go binaries with no crypto symbols.

kolyshkin: Well, here we scan rpms (and ignore everything else), so if there are no rpms and as a result we scanned 0 files we should at least emit a warning.

Maybe we should rename this command from `scan node` to `scan rpmroot`, and add `scan node` command which will do walk over all files (same as we do for `scan payload`. WDYT?

mrunalp: @kolyshkin Yes, something like that may make sense from a naming perspective.

mrunalp (approved):
/lgtm

PR #91, opened 2023-07-26
test: check that embedded configs are sane

kolyshkin:
The test checks that
 - configs can be parsed
 - there are no unknown/unparsed parts
 
 This ensures there are no runtime errors due to bad embedded configs.


mrunalp (approved):
/lgtm

Issue #94, opened 2023-07-27
Check filenames in config


kolyshkin:
We should implement checking that all paths (dirs, files) in configuration files are
 - absolute
 -  cleaned
 
 Otherwise, something might stop working in some unexpected ways (e.g. when a directory is provided with an ending `/`, which looks totally fine but won't work).

Issue #93, opened 2023-07-26
Config merge improvements


kolyshkin:
Currently, the ignore lists from the add-on config are appended to the main one, and if there are duplicates, they are appended, too. This works just fine, but can be improved in the following ways:

 - duplicates should not be added, as they makes things slower;
 - for embedded configs, we can check there are no duplicates, and fail CI otherwise;
 - for user-specified configs, we can warn (rather than error) on duplicates.
 
 Same about ignore lists -- currently they are appended, but the lists with the same error name should be merged into the same key.

PR #95, opened 2023-07-27
Better config merge and validation

kolyshkin:
When merging config files, check for duplicates and warn (during runtime) or error out (in CI) when a dupes are there.

Add some primitive config validation -- check that all the file/dir entries are absolute and cleaned paths, also checks for duplicates.

Add some test cases.

Please review commit-by-commit, and check individual commit messages for details.

Fixes: #94
Fixes: #93


rphillips:
nice 

/lgtm

PR #105, opened 2023-08-04
sort config versions for display

rphillips:


dist/releases/embeds.go
```diff
@@ -4,6 +4,9 @@ import (
 	"embed"
 	"fmt"
 	"path/filepath"
+	"sort"
+
+	"github.com/hashicorp/go-version"
```
kolyshkin, line 9: Since we already use `github.com/Masterminds/semver/v3` perhaps we can also use it here? I checked it has version comparison functions.

rphillips: Ah good call.

rphillips: fixed!

dist/releases/4.14/config.toml
```diff
@@ -0,0 +1,79 @@
...
+files = [ "/usr/bin/skopeo" ]
+
+[[payload.ose-machine-config-operator-container.ignore]]
+error = "ErrLibcryptoSoMissing"
+files = [ "/usr/bin/machine-config-daemon.rhel9" ]
+
+[[rpm.containernetworking-plugins.ignore]]
+error = "ErrGoMissingTag"
+files = [ "/usr/libexec/cni/dummy" ]
+
```
kolyshkin, line 34: Looks like you've copied it from 4.13 config, which I forgot to fix. Please see https://github.com/openshift/check-payload/pull/106

kolyshkin (changes requested):
1. I think we should not use two different packages for dealing with versions.
2. Since a bunch of exceptions are the same/similar for 4.12, 4.13, 4.14, maybe we should move them to the main config? I'm thinking rpm.cri-o, rpm.podman etc.

rphillips:
hmm yeah... @kolyshkin could you PR that, and I can rebase this PR with perhaps just the sorting change.

CHANGELOG.md
```diff
@@ -4,7 +4,8 @@
 
 ### Features
 
-- Add 4.14 configuration.
+- Add 4.14 config file
```
kolyshkin, line 7: why?

kolyshkin:
@rphillips can you squash the commits?

kolyshkin:
Sorry, it seems we clicked on "edit" for title at the same time  😆

kolyshkin (approved):
lgtm

PR #106, opened 2023-08-04
chore: remove a dup from 4.13 config

kolyshkin:
This is already excluded by main config.toml:

> [[rpm.containernetworking-plugins.ignore]]
> error = "ErrGoMissingTag"
> dirs = [ "/usr/libexec/cni" ]


kolyshkin:
This issue has slipped from config validation because of the bug, now fixed in #108

PR #117, opened 2023-09-21
config.toml: add exception for rukpak

kolyshkin:
Joe Lanford says:

> We copy this binary into a scratch image, so it needs to be statically
> compiled. It does not do any crypto.

Since this binary will also be in a future releases, place it in a global config.

Tested with:
```console
$ ./check-payload scan payload -u registry.ci.openshift.org/ocp/release:4.14.0-0.nightly-2023-09-21-002530 -V 4.14
...
---- Successful run
```


mrunalp (approved):
/lgtm

PR #118, opened 2023-09-21
exclude a statically-compiled utility for OLM

stevekuznetsov:
In 4.15+, OLM will not require that users provide an `opm` binary with their index image. In cases where they do not, OLM simply copies the catalog data from their index image and uses it directly. In order to copy this data, we do not want to assume anything about the image that contains the content we need to copy, so we statically compile a utility that helps us copy the data around, instead of assuming the image has `sh` or `cp`.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -77,7 +77,10 @@ error = "ErrNotDynLinked"
 files = [ "/usr/bin/pod" ]
 
 [payload.operator-lifecycle-manager-container]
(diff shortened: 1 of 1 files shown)


stevekuznetsov:
https://github.com/operator-framework/operator-lifecycle-manager/issues/3019

mrunalp (approved):
/lgtm

kolyshkin (approved):
LGTM

PR #119, opened 2023-10-12
Add other fips check gosyms tool to support running pods scan

xiaojiey:
To support fips check for all the binaries in all the running pods, add gosyms tool

### DIFF
--- a/tools/gosyms/gosyms
+++ b/tools/gosyms/gosyms
@@ -1,0 +1,3 @@
+#!/bin/bash -e
+DIR=$(dirname $(readlink -f $0))
+go run  $DIR/gosyms-example.go $@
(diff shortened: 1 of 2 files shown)


xiaojiey:
@rphillips @mrunalp What's your opinion on this PR? Thanks.

rphillips:
@xiaojiey is going to refactor this to be a subcommand.

kolyshkin:
> @xiaojiey is going to refactor this to be a subcommand.

@xiaojiey It would be nice to reuse the code from internal/golang/goscan.go (currently it seems it's a copy-paste, which creates a maintenance burden).

xiaojiey:
/close
I will raise a new PR instead.

PR #152, opened 2024-01-29
x_cgo: check for _cgo_topofstack

rphillips:


mrunalp (commented):
I would suggest adding a commit comment about why we are doing this.

mrunalp (approved):
/lgtm

PR #154, opened 2024-01-31
Add a "local" scan sub-command avoids podman mount

bentito:
This adds a "local" scan option.

Assuming the file structure has already been pulled down and unpacked, for instance:

```sh
$ skopeo copy --override-os=linux --override-arch=amd64 docker://registry.redhat.io/rhscl/redis-5-rhel7@sha256:<snip> oci:bundle-oci:redis
$ sudo umoci unpack --image bundle-oci:redis bundle
$ tree -L 2 bundle
bundle
├── config.json
├── rootfs
│   ├── bin -> usr/bin
│   ├── boot
│   ├── dev
│   ├── etc
│   ├── help.1
│   ├── home
│   ├── lib -> usr/lib
│   ├── lib64 -> usr/lib64
│   ├── media
│   ├── mnt
│   ├── opt
│   ├── proc
│   ├── root
│   ├── run
│   ├── sbin -> usr/sbin
│   ├── srv
│   ├── sys
│   ├── tmp
│   ├── usr
│   └── var
├── sha256_73a93123f9be132f8640fe9bf7fabf903c660dff144815a9f1d58843e23f4489.mtree
└── umoci.json
```
Then this new subcommand allows for scanning the already unpacked image. See README.md diff for usage.


hack/create_mock_bundles.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
```
rphillips, line 2: `set -eou pipefail`

will exit if there is an error.

internal/scan/scan.go
```diff
@@ -125,6 +126,68 @@ func ValidateTag(ctx context.Context, cfg *types.Config, tag *v1.TagReference, r
...
+		result := validateTagLocal(ctx, req.Tag, cfg, localBundlePath)
+		rx <- &Result{Tag: req.Tag, Results: result}
+	}
+}
+
+// simulateLocalPayload generates a slice of v1.TagReference based on the local bundle directory structure.
+// Adjust this to match how your local bundle's tags are represented in the file system if needed
+func simulateLocalPayload(localBundlePath string) []*v1.TagReference {
+	var tags []*v1.TagReference
+	// ... logic to populate tags based on localBundlePath ...
```
rphillips, line 187: Is this function going to be filled out?

internal/scan/scan_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+	// Iterate over test cases
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			// Setup context and config
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+
+			var cfg types.Config
+			cfg.SetupDefault() // Assuming a method to set up default config
```
rphillips, line 31: Add a NewDefaultConfig() function makes sense... Keep the SetupDefault (or perhaps rename it to SetupDefaults()) .

rphillips:
Overrall, this PR looks great. Just a couple minor questions and the verify linter failed.

hack/create_mock_bundles.sh
bentito: thanks for the catch, adding

internal/scan/scan_test.go
bentito: yep, changed it to the former suggestion, better fit for what it does.

bentito:
I added a known good binary and libcrypto.so to the "good" mock bundle in filesystem, so now there is a good and bad path test for the new sub command. @rphillips let me know on these test additions since you looked last and if they're good maybe we're able to merge?

internal/scan/scan.go
bentito: I reworked this function a bit more so it would allow to stay in for the linter, but it does do nothing as-is, just useful maybe for future scaffold.

README.md
```diff
@@ -54,6 +54,34 @@ Here
...
+### Scan a Local Unpacked Image Bundle
+
+The `scan local` subcommand allows you to scan a local unpacked image bundle for FIPS compliance. This is particularly useful for testing or analyzing local images that have been unpacked using tools like `umoci`.
+
+#### Usage
+
+To scan a local unpacked image bundle, use the following command:
+
+```sh
+./check-payload scan local --path /path/to/local/bundle
```
nmars, line 66: Will `scan local --path` scan unpacked operator bundles (which include the `/manifests` and `/metadata` directories), and/or unpacked "operand" images, which are "normal" images with the usual expected directory structure (`/etc/`, `/usr/`, `/sbin` and so forth).

Apologies if I've misunderstood something. I am new to this tool. :)

hack/create_mock_bundles.sh
```diff
@@ -0,0 +1,52 @@
...
+    for SUBDIR in "$@"; do
+        local FULL_PATH="$BUNDLE_DIR/$SUBDIR"
+        mkdir -p "$FULL_PATH"
+        touch "$FULL_PATH/.gitkeep"
+        echo "Created $FULL_PATH with .gitkeep"
+    done
+}
+
+# Create the bundle directories and subdirectories
+create_bundle "bundle-1" "etc" "usr" "usr/lib64"
```
nmars, line 35: This is maybe a nit, but this seems to create a mock unpacked image with /etc/, /usr, /usr/lib64. 

Use of the term "bundle" confused me a bit because I think of https://github.com/operator-framework/operator-registry/blob/master/docs/design/operator-bundle.md

which is a stripped down image whose only content is metadata in the /manifests and /metadata directories (and maybe a couple other directories).

bentito: yeah I can see that confusion. changing the naming to "mock unpacked filesystem" in d21f5cdc

README.md
bentito: `check-payload scan local` should scan any unpacked file system, it will walk the filesystem from the provided "--path" and should work as it would after the (skipped) `podman mount` phase in the `scan operator` work flow.

rphillips:
Awesome. Thank you!

/lgtm
/approve

internal/types/types.go
```diff
@@ -131,3 +131,36 @@ func (ve *ValidationError) IsError() bool {
 func (ve *ValidationError) IsWarning() bool {
 	return ve.Level == Warning
 }
+
+// NewDefaultConfig initializes the configuration with default values for testing.
+func (c *Config) NewDefaultConfig() {
```
kolyshkin, line 136: 1. If that's for testing, it should be in `_test.go` file.
2. There is no need to set default values (like empty strings, `false` for booleans etc).
3. The name `Default` is misleading -- there is no "default" config.

Perhaps we can have a testing function named `getEmptyConfig` or somesuch.

Or, fix the code that uses the config to allow a truly default config (i.e. `cfg := types.Config{}` should work). If you are going to do that, please do it in a separate commit.

bentito: Seem like reasonable changes, but maybe for a separate PR now?

PR #168, opened 2024-04-08
Add exception for bond cni RHEL 8 binary in RHEL 9 base

mrunalp:
https://github.com/openshift/bond-cni/blob/release-4.16/Dockerfile.openshift copies RHEL 8 binary to RHEL 9 base


kolyshkin (approved):
/lgtm

PR #175, opened 2024-04-24
Implement an operating system check for FIPS certified distributions

rhmdnd:
Some operating systems have gone through the certification process, and
we can use that information to let users know if they're using a version
or operating system that hasn't been FIPS certified, yet.


rhmdnd:
This patch still needs a lot of work, but I wanted to at least sketch out a quick approach based on what I've read in the code so far.

internal/validations/validations_os.go
```diff
@@ -0,0 +1,43 @@
...
+	f, err := os.Open(path)
+	if err != nil {
+		e := errors.New("release file not found")
+		info.Error = e
+		return info
+	}
+	defer f.Close()
+
+	certifiedReleases := []string{
+		"Red Hat Enterprise Linux release 9.2 (Plow)",
```
mrunalp, line 30: We should move this to config so we can specify certified OS per release.
For 4.16 it will be 9.2 etc.

rphillips: Great idea.

internal/validations/validations_os.go
```diff
@@ -0,0 +1,43 @@
...
+		Certified: false,
+		Error:     nil,
+	}
+
+	path := filepath.Join(mountPath, "/etc/redhat-release")
+	info.Path = path
+	f, err := os.Open(path)
+	if err != nil {
+		e := errors.New("release file not found")
+		info.Error = e
```
kolyshkin, line 24: nit: why do you need an intermediate variable (`e`) here?

kolyshkin: Also, it makes sense to return the actual error here (which can help to figure out what is going on).

internal/validations/validations_os.go
rhmdnd: We expect to have multiple certified versions per OpenShift release, right?

hack/create_mock_bundles.sh
```diff
@@ -40,6 +40,7 @@ echo "Copying binaries to mock_unpacked_dir-1..."
 cp "$TEST_RESOURCES_DIR/fips_compliant_app" "$TEST_RESOURCES_DIR/mock_unpacked_dir-1/usr/fips_compliant_app"
 cp "$TEST_RESOURCES_DIR/libcrypto.so" "$TEST_RESOURCES_DIR/mock_unpacked_dir-1/usr/lib64/libcrypto.so"
 cp "$TEST_RESOURCES_DIR/libcrypto.so" "$TEST_RESOURCES_DIR/mock_unpacked_dir-1/usr/lib64/libcrypto.so.1.1"
+cp "$TEST_RESOURCES_DIR/redhat-release" "$TEST_RESOURCES_DIR/mock_unpacked_dir-1/etc/redhat-release"
```
rhmdnd, line 43: This technically isn't a binary - so the output above isn't exactly accurate...

config.toml
```diff
@@ -49,6 +49,10 @@ java_fips_disabled_algorithms = [
   "HmacMD5",
 ]
 
+certified_distributions = [
```
rhmdnd, line 52: Pending feedback on this approach, we can start proposing these changes to various OpenShift version configs (e.g., 4.16)

internal/types/types_scan_result.go
```diff
@@ -57,6 +57,13 @@ func (r *ScanResult) SetOpenssl(info OpensslInfo) *ScanResult {
 	return r
 }
 
+func (r *ScanResult) SetOperatingSystem(info OperatingSystemInfo) *ScanResult {
+	if !info.Certified {
+		r.SetError(errors.New("operating system is not FIPS certified"))
```
mrunalp, line 62: Let us introduce a new error for this in internal/types/error.go

internal/validations/validations_os.go
mrunalp: For 4.16, the base images will be 9.2 except for the node itself which will be on 9.4. So, yes, it will be good to make this a list to support multiple versions.

config.toml
rphillips: Thinking this might be better in the versioned configs, since different releases will be based on different base operating systems.

mrunalp: Yes, agree :+1:

rhmdnd: Ack - working that feedback into this approach.

rhmdnd: The code to find the configuration value works with both this config and the version-specific configs. The only reason why this is still here is because the scan tests use this config in their testing.

test/resources/mock_unpacked_dir-1/usr/lib64/libcrypto.so.1.1
```diff
@@ -1 +1 @@
-hack/../test/resources/mock_unpacked_dir-1/usr/lib64/libcrypto.so
\ No newline at end of file
+./hack/../test/resources/mock_unpacked_dir-1/usr/lib64/libcrypto.so
```
rhmdnd, line 1: Looks unrelated... reverting.

dist/releases/4.16/config.toml
```diff
@@ -1,3 +1,7 @@
+certified_distributions = [
+  "Red Hat Enterprise Linux release 9.2 (Plow)",
+]
+
```
kolyshkin, line 4: Version-specific configuration is an add-on to system-wide one (IOW version-specific one is merged into the global one). Here you add the same string to both global and version-specific configs, which is redundant.

Having said that, what might be missing is the merging code (haven't checked).

kolyshkin: In other words, if something is in the global config, it does not have to be repeated.

internal/types/errors.go
```diff
@@ -18,4 +18,5 @@ var (
 	ErrLibcryptoMissing   = errors.New("openssl: did not find libcrypto library within binary")
 	ErrLibcryptoSoMissing = errors.New("could not find dependent openssl version within container image")
 	ErrNotDynLinked       = errors.New("executable is not dynamically linked")
+	ErrOsNotCertified     = errors.New("operating system is not FIPS certified")
```
kolyshkin, line 21: nit: this should be `ErrOSNotCertified`.

dist/releases/4.16/config.toml
kolyshkin: For this to work, you need to amend `func (c *ConfigFile) Add(add *ConfigFile) error` using `appendUniq` (and maybe add appropriate test cases).

internal/scan/scan_test.go
```diff
@@ -28,6 +30,10 @@ func TestRunLocalScan(t *testing.T) {
 
 			var cfg types.Config
 			cfg.NewDefaultConfig() // Assuming a method to set up default config
+			_, err := toml.DecodeFile("../../config.toml", &cfg)
+			if err != nil {
+				t.Errorf("Failed to parse configuration file: %s", err)
+			}
```
kolyshkin, line 36: Why is this added here? We do test config files elsewhere (see `dist/releases/config_test.go`).

internal/validations/validations_os.go
```diff
@@ -0,0 +1,43 @@
...
+
+	scanner := bufio.NewScanner(f)
+	for scanner.Scan() {
+		for _, release := range certifiedReleases {
+			if strings.Contains(scanner.Text(), release) {
+				info.Certified = true
+				break
+			}
+		}
+	}
```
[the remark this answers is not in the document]
rhmdnd: Done - thanks for the suggestions.

internal/scan/scan_test.go
rhmdnd: This is a byproduct of not having a default value for certified distributions.

It appears the configuration we're using in the test, which is initialized on line 32, is using a set of reasonable defaults:

  https://github.com/openshift/check-payload/blob/main/internal/types/types.go#L137-L165

But, the certified distributions are empty.

Should we add in a default value instead?

dist/releases/4.16/config.toml
rhmdnd: Nice -thanks for the recommendations. I worked that into the latest revision.

internal/scan/scan_test.go
kolyshkin: > Should we add in a default value instead?

Maybe.

Also, as I mentioned [here](https://github.com/openshift/check-payload/pull/154#discussion_r1480835381), the NewDefaultConfig should be in a `_test.go` file since we're only using it for testing.

Maybe we can even remove `NewDefaultConfig` entirely (as the name is misleading, and it is only used by a single test case), and have some config right there in `TestRunLocalScan`.

Opened https://github.com/openshift/check-payload/pull/191

internal/validations/validations_os.go
```diff
@@ -0,0 +1,35 @@
...
+
+	"github.com/openshift/check-payload/internal/types"
+)
+
+const releaseFilePath = "/etc/redhat-release"
+
+func ValidateOperatingSystem(cfg *types.Config, mountPath string) types.OperatingSystemInfo {
+	info := types.OperatingSystemInfo{
+		Certified: false,
+		Error:     nil,
```
kolyshkin, line 16: those are both default values, so maybe just 
```go
info := types.OperatingSystemdInfo{}
```

or even

```go
func ValidateOperatingSystem(cfg *types.Config, mountPath string) (info types.OperatingSystemInfo) {
```

kolyshkin (commented):
As a general comment, I'd `s/OperatingSystem/OS/g` everywhere in the code.

internal/scan/scan_test.go
rhmdnd: Rebased on top of your patch after it merged and remove the modifications to `config.toml` by adding it to the testing configuration directly.

internal/validations/validations_os.go
```diff
@@ -0,0 +1,30 @@
...
+
+const releaseFilePath = "/etc/redhat-release"
+
+func ValidateOS(cfg *types.Config, mountPath string) (info types.OSInfo) {
+	certifiedReleases := cfg.GetCertifiedReleases()
+	path := filepath.Join(mountPath, releaseFilePath)
+	info.Path = path
+	f, err := os.ReadFile(path)
+	if err != nil {
+		info.Error = err
```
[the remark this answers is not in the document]
rhmdnd: No - it doesn't look like it. I could remove it and just keep the certified key, unless you think having the error message bubble up somewhere would be useful.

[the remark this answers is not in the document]
rhmdnd: I removed setting the error on info and just logged it instead. It looks like we take that approach else where in the code when attempting to write files.

internal/scan/scan_test.go
```diff
@@ -38,7 +40,7 @@ func TestRunLocalScan(t *testing.T) {
 			ctx, cancel := context.WithCancel(context.Background())
 			defer cancel()
 
-			// Run the local scan.
+			// Run the local scan
```
kolyshkin, line 43: nit: please remove this hunk

internal/types/types_scan_result.go
```diff
@@ -57,6 +57,15 @@ func (r *ScanResult) SetOpenssl(info OpensslInfo) *ScanResult {
 	return r
 }
 
+func (r *ScanResult) SetOS(info OSInfo) *ScanResult {
+	if info.Error != nil {
+		r.SetError(ErrDistributionFileMissing)
```
kolyshkin, line 62: This kind of hides the underlying error. If it is at least logged, it would be OK, but it is not.

I think some images doesn't have `/etc/redhat-release` and maybe they need to be excluded for this error. WDYT @mrunalp ?

rhmdnd: Yeah - I think that's probably case with scratch images. Curious to know if they should be excluded or if we just propose they're not used at all?

mrunalp: Yes, we need to support exclusions for this type of error like we have for other error types. 
Also, we should include that exclusion for the ubi9 and other known failures as part of merging this PR so we don't break existing CI flows.

internal/scan/scan.go
```diff
@@ -358,6 +358,21 @@ func validateTagLocal(ctx context.Context, tag *v1.TagReference, cfg *types.Conf
...
+	if component != nil && osScanResult.Error != nil {
+		if i, ok := cfg.PayloadIgnores[component.Component]; ok {
+			if tag != nil {
+				if !i.ErrIgnores.IgnoreTag(tag.Name, osScanResult.Error.Error) {
+					results.Append(osScanResult)
+				}
+			}
+		}
+	}
+
```
kolyshkin, line 375: So, if something is certified, it is not reflected in the results? IOW only errors are shown?

internal/types/types_config_file.go
```diff
@@ -94,7 +94,7 @@ func validateErrIgnores(section string, perr, pwarn *error, l ErrIgnoreList) {
 			multierr.AppendInto(perr, &errEmpty{section, "error="})
 		}
 		// Make sure files/dirs are not empty.
```
kolyshkin, line 96: Shouldn't this be `files/dirs/tags`?

internal/types/types_config_file.go
```diff
@@ -94,7 +94,7 @@ func validateErrIgnores(section string, perr, pwarn *error, l ErrIgnoreList) {
...
-		if len(v.Files)+len(v.Dirs) == 0 {
+		if len(v.Files)+len(v.Dirs)+len(v.Tags) == 0 {
 			multierr.AppendInto(perr, &errEmpty{section, "files= nor dirs="})
```
kolyshkin, line 98: `files= nor dirs= nor tags=`?

internal/types/types.go
```diff
@@ -48,6 +49,7 @@ type ErrIgnore struct {
 	Error KnownError `toml:"error"`
 	Files []string   `toml:"files"`
 	Dirs  []string   `toml:"dirs"`
+	Tags  []string   `toml:"tags"`
```
kolyshkin, line 52: Looks like this adds an ability to ignore a particular error for a particular tag. If yes, does it work for other checks?

internal/types/types_scan_result.go
kolyshkin: The issue that we still lose the underlying error from `os.ReadFile()` is still not solved.

internal/scan/scan.go
rhmdnd: >  IOW only errors are shown?

Yes

internal/types/types_scan_result.go
rhmdnd: For this are you looking for a log message, or something added directly to the error we pass through in the report?

internal/types/types.go
rhmdnd: No - I believe it's used exclusively for checking the distribution.

kolyshkin: Maybe add a comment about that, as it is counter-intuitive.

internal/scan/scan.go
kolyshkin: This is not the way the rest of the system works. We are able to look into full logs and see that such and such is verified.

@rphillips WDYT?

internal/types/types_scan_result.go
kolyshkin: Since `open` or `read` may fail for many reasons, I guess that we should only translate ENOENT (aka `os.ErrNotExist`) into `os.ErrDistributionFileMissing`, for the sake of being able to exclude it, and report all other errors (as they are not expected) as-is.

kolyshkin: By "report" I mean "add it to scan results".

internal/scan/scan.go
rhmdnd: Fixed.

kolyshkin (approved):
lgtm

PR #178, opened 2024-04-29
Manually update oc to a new version (48dcf5980a)

rhmdnd:
This docker dependency update is failing because the version dependabot is upgrading to doesn't have a function that openshift/oc expects it to have because it was deprecated and removed.

Let's upgrade openshift/oc while we're at it so that it doesn't rely on deprecated and removed functionality of docker libraries.


mrunalp (approved):
/lgtm

PR #183, opened 2024-05-17
Fix braces

dominikholler:
The single "[" resulted in ignoring the following lines in the
config file.

In addition, payload.operator-lifecycle-manager-container and
payload.ose-olm-rukpak-container are missing the error to
mask.


dominikholler:
@kolyshkin @rphillips can you please review

PR #184, opened 2024-05-17
add back rukpak and lifecycle manager

rphillips:


dist/releases/4.16/config.toml
```diff
@@ -387,3 +387,27 @@ files = ["/usr/bin/machine-config-daemon.rhel8"]
 [[payload.sriov-cni-container.ignore]]
 error = "ErrLibcryptoSoMissing"
 files = ["/usr/bin/rhel8/sriov"]
+
+[[payload.operator-lifecycle-manager-container.ignore]]
+error = "ErrGoNotCgoEnabled"
+files = ["/usr/bin/cpb", "/usr/bin/copy-content"]
+
```
kolyshkin, line 394: This seems a duplicate of an entry in the global config.

ashwindasr: +1

Raised a [PR](https://github.com/rphillips/check-payload/pull/1) to this PR

PR #189, opened 2024-05-23
Add skopeo and umoci to CI image

swghosh:
Adds skopeo and umoci binaries to the check-payload image (useful when using `scan local`).

xref: https://redhat-internal.slack.com/archives/C05U13J3LLS/p1716476010936559?thread_ts=1715954957.336589&cid=C05U13J3LLS

### DIFF
--- a/Dockerfile.upstream
+++ b/Dockerfile.upstream
@@ -9,14 +9,18 @@ RUN make
 
 FROM fedora:38
 ARG OC_VERSION=latest
+ARG UMOCI_VERSION=latest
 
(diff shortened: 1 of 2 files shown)


Dockerfile.upstream
```diff
@@ -9,14 +9,18 @@ RUN make
 
 FROM fedora:38
```
mrunalp, line 10: Not blocking but we should probably update this as well.

mrunalp (approved):
/lgtm

PR #201, opened 2024-06-25
set info.Certified as false, if empty string

ashwindasr:
Follows https://github.com/openshift/check-payload/pull/175

`openshift-enterprise-pod-container` was passing scan if `-V 4.16` was set, even though it was not mentioned in the [excludes](https://github.com/openshift/check-payload/blob/main/dist/releases/4.16/config.toml) 

thead [here](https://redhat-internal.slack.com/archives/CB95J6R4N/p1719324150710139?thread_ts=1719309951.266479&amp;cid=CB95J6R4N)

### DIFF
--- a/internal/validations/validations_os.go
+++ b/internal/validations/validations_os.go
@@ -22,8 +22,9 @@ func ValidateOS(cfg *types.Config, mountPath string) (info types.OSInfo) {
 	}
 
 	for _, d := range cd {
-	    cd_bytes = []byte(d)
-		if f != "" && cd_bytes != "" && bytes.HasPrefix(f, cd_bytes) {
+	    cd_bytes := []byte(d)
+
+		if len(f) != 0 && len(cd_bytes) != 0 && bytes.HasPrefix(f, cd_bytes) {
 			info.Certified = true
 			break
 		}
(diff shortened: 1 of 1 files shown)


internal/validations/validations_os.go
```diff
@@ -22,7 +22,8 @@ func ValidateOS(cfg *types.Config, mountPath string) (info types.OSInfo) {
 	}
 
 	for _, d := range cd {
-		if bytes.HasPrefix(f, []byte(d)) {
+	    cd_bytes = []byte(d)
```
rphillips, line 25: `cd_bytes := []byte(d)`

internal/validations/validations_os.go
```diff
@@ -22,7 +22,8 @@ func ValidateOS(cfg *types.Config, mountPath string) (info types.OSInfo) {
...
+		if f != "" && cd_bytes != "" && bytes.HasPrefix(f, cd_bytes) {
```
rphillips, line 26: len(f) != 0 && len(cd_bytes) != 0 && bytes.HasPrefix(f, cd_bytes) ....

ashwindasr:
/hold

The error might be in a different place. 
`sudo ./check-payload -V 4.16 scan image --spec registry-proxy.engineering.redhat.com/rh-osbs/openshift-ose-pod-rhel9@sha256:856bcc5a915f5af7dd6bfb35faa54ecb78e350c4ca5adb01d49df8915d3bb258
` still passes when it should have been failing

ashwindasr:
/close 

In favor of https://github.com/openshift/check-payload/pull/202

PR #202, opened 2024-06-25
Fix OS scan when -V is not set

rphillips:
This PR warns when the os cannot be figured out without the -V option. The default config does not have a valid OS distribution set, since it is version specific.


ashwindasr:
`cd` (certified distributions) was returning an empty list `[]` if `-V` is not set. This PR aims to fix that by failing with a new error (`certified_distributions is not defined, please use -V [VERSION] on the command line`). For example:

```
| openshift-enterprise-pod-container | /etc/redhat-release | certified_distributions is not defined, please use -V [VERSION] on the command line | registry-proxy.engineering.redhat.com/rh-osbs/openshift-ose-pod-rhel9@sha256:856bcc5a915f5af7dd6bfb35faa54ecb78e350c4ca5adb01d49df8915d3bb258 |
```

internal/validations/validations_os.go
```diff
@@ -22,7 +22,8 @@ func ValidateOS(cfg *types.Config, mountPath string) (info types.OSInfo) {
 	}
 
 	for _, d := range cd {
-		if bytes.HasPrefix(f, []byte(d)) {
+	    cd_bytes = []byte(d)
```
kolyshkin: you have spaces instead of a tab here.

ashwindasr:
also looks like verify is failing, we may need to add 
```
"ErrCertifiedDistributionsEmpty": ErrCertifiedDistributionsEmpty,
```
to `internal/types/error_map.go`

kolyshkin:
Looks like the first two commits needs to be squashed?

internal/types/types_scan_result.go
```diff
@@ -59,10 +58,8 @@ func (r *ScanResult) SetOpenssl(info OpensslInfo) *ScanResult {
 }
 
 func (r *ScanResult) SetOS(info OSInfo) *ScanResult {
-	if errors.Is(info.Error, os.ErrNotExist) {
-		r.SetError(ErrDistributionFileMissing)
```
rhmdnd, line 63: These were the only two places using `ErrDistributionFileMissing`. We can probably consider removing the error from errors.go.

rphillips: I added this back in within a different spot. Thanks.

rphillips:
Squashed and fixups the comments.

PR #204, opened 2024-06-27
bump deps

rphillips:
The mergo import path changed, so there is now a replace for it. This is why dependabot wasn't working.


mrunalp (approved):
/lgtm

PR #205, opened 2024-06-28
CMP-2639: Exclude binary libcrypto check using golang 1.22

rhmdnd:
Golang version 1.22 does not have hardcoded libcrypto versions anymore, and does not emit the libcrypto version in the binary output of a go binary (e.g., using strings).

Because of this, the approach we were using to detect the OpenSSL version based on a regular expression and attempting to match that with the binary output does not work for binaries built with golang 1.22.

This is particularly prevalent with OpenShift 4.17 payloads that are using RHEL 9 and golang 1.22 base images.

This commit updates the binary scan logic to skip checking binaries for OpenSSL versions when using golang 1.22.

### DIFF
--- a/internal/validations/validations.go
+++ b/internal/validations/validations.go
@@ -38,8 +38,9 @@ var (
 		"vendor/github.com/golang-fips/openssl/v2.dlopen",
 	}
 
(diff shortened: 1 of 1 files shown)


mrunalp (approved):
/lgtm

PR #206, opened 2024-07-03
Pass component information to scan local check

### LINKED ISSUE #193: Payload exception rules not matched by "scan local" command
`check-payload scan local` is not able to match payload* exception rules as entered in `config.toml` or the toml files under `dist/releases`.

```
$ check-payload version
0.3.1-140-g8d062caf
```

Using the `scan image` command that relies on podman does work:

```
$ sudo check-payload scan -V 4.16 image --spec registry.redhat.io/rhacm2/volsync-rhel9@sha256:a9dbd64e00caebf06ed38708fd5464219ccd44132f717e73d806165c5a16d050
...
---- Successful run
```

However, a local scan, where the rootfs has been unpacked to /tmp fails:

```
$ check-payload scan -V 4.16 local --path /tmp/rhacm-volsync-rhel9/unpacked-rootfs/
...
I0610 15:37:42.807182   22948 scan.go:429] "scanning failed" image="" path="/usr/local/bin/diskrsync" error="go binary does not contain required symbol(s)" component="" tag="" rpm="" status="failed"
---- Failure Report
+--------------------------+-----------------------------------------------+
| EXECUTABLE NAME          | STATUS                                        |
+--------------------------+-----------------------------------------------+
| /usr/local/bin/diskrsync | go binary does not contain required symbol(s) |
+--------------------------+-----------------------------------------------+
```

The local scan does not match the rules in the exception file:

```
$ grep -B5 /usr/local/bin/diskrsync dist/releases/4.16/config.toml
# VolSync packages diskrsync which uses x/crypto/blake2b for local hashing only
# for comparing blocks of data (non-cryptographic)
# Actual network transfer is handled by the ssh executable in the image
[[payload.volsync-container.ignore]]
error = "ErrGoMissingSymbols"
files = ["/usr/local/bin/diskrsync"]

[[payload.volsync-container.ignore]]
error = "ErrNotDynLinked"
files = ["/usr/local/bin/diskrsync"]

[[payload.volsync-container.ignore]]
error = "ErrLibcryptoMissing"
files = ["/usr/local/bin/diskrsync"]
```

I think this is due to the `component` not being discovered from the `com.redhat.component` label during a local scan. 

If I'm right, maybe the payload could be discovered via `skopeo inspect` instead of podman. Or maybe a new flag can be created for `scan local` so users can specify what component to use to look for the exception rules.

rhmdnd:
Previously, we weren't passing in the `OpenshiftComponent` to the local
scan check. This meant that when you ran a local scan against the file
system of a container mounted locally, check-payload wouldn't honor the
exclusions implemented in the global configuration or the versioned
configuration files available in the check-payload binary.

This commit updates the scan local command to pass this information in,
and ports the component string argument to an `OpenshiftComponent`
struct, allowing us to reuse the existing exclusions. Ultimately making
it so `scan local` and `scan payload` behave similarly.

Fixes #193

### DIFF
--- a/internal/scan/scan.go
+++ b/internal/scan/scan.go
@@ -348,11 +348,22 @@ func validateTagLocal(ctx context.Context, tag *v1.TagReference, cfg *types.Conf
 		return types.NewScanResults().Append(types.NewScanResult().SetTag(tag).SetError(err))
 	}
 
-	// Since the local bundle does not require a pull or mount, we skip directly to scanning.
-	// Assume OpenshiftComponent information is either not needed or can be derived
-	// from the local bundle structure. If needed, create a mock or derive it here.
+	// Since the local bundle does not require a pull or mount, we skip
+	// directly to scanning. In this case, OpenshiftComponent information
+	// is either not supplied, or a list of one that maps to the unpacked
+	// image from the local bundle structure. At this point, validation
+	// logic will have prevented users from passing in more than one
+	// component.
+	component := &types.OpenshiftComponent{}
+	switch len(cfg.Components) {
+	case 0:
+		// Scan without any component filtering to maintain backwards
+		// compatibility.
+	case 1:
+		component.Component = cfg.Components[0]
+	}
 
-	return walkDirScan(ctx, cfg, tag, nil, localTagPath)
+	return walkDirScan(ctx, cfg, tag, component, localTagPath)
 }
 
 func walkDirScan(ctx context.Context, cfg *types.Config, tag *v1.TagReference, component *types.OpenshiftComponent, mountPath string) *types.ScanResults {
--- a/main.go
+++ b/main.go
@@ -99,6 +99,7 @@ func main() {
 			config.TimeLimit = timeLimit
 			config.Verbose = verbose
 			config.Log()
+			config.Components = components
 			klog.InfoS("scan", "version", Commit)
 
 			// Validate the configuration.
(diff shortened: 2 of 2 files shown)


internal/scan/scan.go
```diff
@@ -351,8 +351,12 @@ func validateTagLocal(ctx context.Context, tag *v1.TagReference, cfg *types.Conf
 	// Since the local bundle does not require a pull or mount, we skip directly to scanning.
 	// Assume OpenshiftComponent information is either not needed or can be derived
 	// from the local bundle structure. If needed, create a mock or derive it here.
+	component := &types.OpenshiftComponent{}
+	for _, c := range cfg.Components {
+		component.Component = c
```
rhmdnd, line 356: This only works with a list of one... we either need to make sure that's always the case when invoked in this way, or call `walkDirScan` for each component in the list.

@mrunalp @rphillips @kolyshkin what are your opinions here?

mrunalp: I think a single component should suffice.

kolyshkin: 1. This code was originally added in #154, and the author (@bentito) added a comment below saying component info is not needed. At the very least, this comment needs to be updated to reflect the reality. Perhaps @bentito can suggest something.

2. The range statement used here gives a false sense that we do consume all the components, while we only use the last one. This is wrong. There should at least be a warning saying we use one component and discarding the rest.

Maybe do something like this (NOTE this code is using the first rather than the last component; this can be changed if needed by using len instead of 0):

```go
component := &types.OpenshiftComponent{}
switch len(cfg.Components) {
   case 0:
        // Do nothing.
    case 1:
        component.Component := cfg.Components[0]
   default: // More than 1.
        component.Component := cfg.Components[0]
        klog.Warning("Multiple components (%d) provided; only using the first one (%q) for the scan",
            len(cfg.Components), component.Component)
}
}
```

main.go
```diff
@@ -188,6 +189,18 @@ func main() {
...
+			// local scan, we're dealing with too much ambiguity.
+			// Otherwise, the local scan needs to run for each
+			// component supplied, when it only maps to a single
+			// component. In this case, the user needs to either
+			// use an empty list (no component filtering) or reduce
+			// their filter to a list of one to remove ambiguity
+			// and allow check payload to safely determine the
+			// component to filter against.
+			if len(components) > 1 {
+				return fmt.Errorf("local scans do not support multiple components, use one or none: %s", components)
```
rhmdnd, line 202: This is a slight departure from what @kolyshkin suggested, but depends on how strict we want to be in allowing users to pass in arguments we won't use.

kolyshkin: Fine with me.

internal/scan/scan.go
rhmdnd: Is that something we want to build out here? And if so, do you want that proposed in this patch or separately?

kolyshkin: Some basic tests would be nice to have, just to make sure we don't break anything. And yes, I think it's out of scope for this PR.

kolyshkin (approved):
lgtm

PR #212, opened 2024-08-05
OCPBUGS-37846: rpm.runc ignore ErrLibcryptoMissing

sdodson:
The change to not link against openssl was backported to 4.16-4.12
as part of OCPBUGS-36541 but seems to manifest with different error
there than in 4.17 due to differences in golang versions.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -76,6 +76,11 @@ files = ["/usr/bin/runc"]
 error = "ErrGoMissingSymbols"
 files = ["/usr/bin/runc"]
 
+[[rpm.runc.ignore]]
+# See OCPBUGS-36541.
+error = "ErrLibcryptoMissing"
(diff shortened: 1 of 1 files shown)


config.toml
```diff
@@ -76,6 +76,11 @@ files = ["/usr/bin/runc"]
 error = "ErrGoMissingSymbols"
 files = ["/usr/bin/runc"]
 
+[[rpm.runc.ignore]]
+# See OCPBUGS-36541.
+error = "ErrLibcryptoSoMissing"
```
kolyshkin, line 81: This should be `ErrLibcryptoMissing` (which is a different one from `ErrLibcryptoSoMissing`, see https://github.com/openshift/check-payload/blob/main/internal/types/errors.go)

kolyshkin (approved):
lgtm

PR #213, opened 2024-08-08
openshift-enterprise-operator-sdk-container update exclude

ashwindasr:
ref slack [thread](https://redhat-internal.slack.com/archives/CB95J6R4N/p1723054496104479)

PR #214, opened 2024-08-20
openshift-enterprise-operator-sdk-container update exclude

ashwindasr:
Follows https://github.com/openshift/check-payload/pull/213

Dockerfile: https://github.com/openshift/ocp-release-operator-sdk/blob/release-4.17/release/sdk/Dockerfile

In the initial Slack [thread](https://redhat-internal.slack.com/archives/CB95J6R4N/p1723054496104479) we only saw failures in 4.13-4.15, but now it seems to be failing in 4.17 and 4.18 as well. 

No issues reported in 4.16 though, not sure why.


rhmdnd:
/lgtm

Strange... thanks for the update @ashwindasr

PR #215, opened 2024-08-22
multicluster-operators-subscription-operator-container excludes for ACM

gparvin:
ACM 2.12 build has a check-payload failure for an extra binary we include compiled on rhel8. Error link:
https://jenkins-cvp-5e54f0365a134668768b8f38.apps.ocp-c1.prod.psi.redhat.com/job/cvp-redhat-operator-bundle-image-validation-test/400/artifact/operator-fips-static-check-bundle-image-output.txt

### DIFF
--- a/dist/releases/4.12/config.toml
+++ b/dist/releases/4.12/config.toml
@@ -37,3 +37,7 @@ files = ["/usr/bin/cpb"]
 [[payload.operator-lifecycle-manager-container.ignore]]
 error = "ErrLibcryptoMissing"
 files = ["/usr/bin/cpb"]
+
+[[payload.multicluster-operators-subscription-operator-container.ignore]]
+error = "ErrLibcryptoSoMissing"
+files = ["/policy-generator/PolicyGenerator-rhel8"]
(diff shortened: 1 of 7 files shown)


dist/releases/4.12/config.toml
```diff
@@ -37,3 +37,7 @@ files = ["/usr/bin/cpb"]
 [[payload.operator-lifecycle-manager-container.ignore]]
 error = "ErrLibcryptoMissing"
 files = ["/usr/bin/cpb"]
+
+[[payload.multicluster-operators-subscription-operator-container.ignore]]
+error = "ErrLibcryptoSoMissing"
+files = ["/policy-generator/PolicyGenerator-rhel8"]
```
rhmdnd, line 43: @gparvin do you have a link to where or how this container is built?

Does is do any sort of cryptographic operations?

gparvin: We build the flagged binary with the OpenShift golang builder for RHEL 8: `openshift-golang-builder:v1.21.11-202407021847.g1ac3e39.el8` so that it will be FIPS ready if it is copied to a RHEL 8 FIPS enabled container.  
Since this binary is not intended for use in the multicluster subscription operator RHEL 9 container we get the error shown.  The build environment uses CPaaS so i can send you that info if you really need it.  The repository the policy-generator is built from is: https://github.com/stolostron/policy-generator-plugin.
Thanks

PR #216, opened 2024-09-05
CMP-2734: Detect usage of x/crypto

rhmdnd:
The x/crypto module isn't compatible with FIPS. Let's write a check to
see if it is in the symbols for a given binary. Ideally, this should
help us find areas where we can remove x/crypto usage altogether.

### DIFF
--- a/internal/validations/validations.go
+++ b/internal/validations/validations.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"regexp"
 	"strings"
@@ -67,6 +68,7 @@ var validationFns = map[string][]ValidationFn{
 		validateGoStatic,
 		validateGoOpenssl,
 		validateGoTagsAndExperiment,
+		validateExcludedCryptoModules,
 	},
 	"exe": {
 		validateNotStatic,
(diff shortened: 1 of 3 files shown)


rhmdnd:
/hold

Holding for some feedback.

internal/validations/validations.go
```diff
@@ -288,6 +290,23 @@ func validateNotStatic(_ context.Context, _ string, baton *Baton) *types.Validat
...
+	var stdout bytes.Buffer
+	cmd := exec.CommandContext(ctx, "nm", "-j", path)
+	cmd.Stdout = &stdout
+	if err := cmd.Run(); err != nil {
+		return types.NewValidationError(err)
+	}
+
+	// Make this more flexible by deriving the excluded modules from
+	// configuration.
+	excluded := []byte("golang.org/x/crypto")
```
rhmdnd, line 303: Ideally - this could be a two-dimensional array that we can load from configuration.

However, that will mean we need to duplicated the excluded modules for each release.

We could invert the logic to opt-into excluding x/crypto by default, and having a configuration option to opt out.

rhmdnd:
@yuumasato in case you're curious how this is shaping up.

internal/validations/validations.go
```diff
@@ -288,6 +290,23 @@ func validateNotStatic(_ context.Context, _ string, baton *Baton) *types.Validat
 	return types.NewValidationError(types.ErrNotDynLinked)
 }
 
+func validateExcludedCryptoModules(ctx context.Context, path string, _ *Baton) *types.ValidationError {
+	var stdout bytes.Buffer
+	cmd := exec.CommandContext(ctx, "nm", "-j", path)
```
dbenoit17, line 295: It was found in the initial scanner implementation that object level symbols are not reliable enough, since they can be stripped from the binary.  Because of this, function data is read by the scanner from the Go runtime's PCLN table instead.  See [validateGoSymbols](https://github.com/openshift/check-payload/blob/d67cb51bcbfb5c115582f6bf46d8c65c6c14b3fe/internal/validations/validations.go#L76-L99).

In the case of the current validation, something like this could probably do the trick:
```
var symbols []string
excluded := "golang.org/x/crypto"
symtable, err := golang.ReadTable(path, baton.GoBuildInfo)
if err != nil {
	return types.NewValidationError(fmt.Errorf("go: could not read table for %v: %w", filepath.Base(path), err))
}
for _, f := range symtable.Funcs {
	if strings.Contains(f.Name, excluded) {
		return types.NewValidationError(types.ErrDetectedExcludedModule)
	}
}

```

dbenoit17 (changes requested):

internal/validations/validations.go
```diff
@@ -288,6 +290,29 @@ func validateNotStatic(_ context.Context, _ string, baton *Baton) *types.Validat
...
+	if err := cmd.Run(); err != nil {
+		return types.NewValidationError(err)
+	}
+
+	// Make this more flexible by deriving the excluded modules from
+	// configuration.
+	excluded := []byte("golang.org/x/crypto")
+	symtable, err := golang.ReadTable(path, baton.GoBuildInfo)
+	if err != nil {
+		return types.NewValidationError(fmt.Errorf("go: could not read table for %v: %w", filepath.Base(path), err))
```
rhmdnd, line 306: We should consider using a warning here, since not all usage of x/crypto violates FIPS.

PR #220, opened 2024-09-22
Add RHEL 9.4 to certified distributions

sdodson:
RHEL 9.2 and 9.4 share openssl fips providers and thus should be equivalent


sdodson:
@rhmdnd We just moved 4.16 to use 9.4 repos which aligns with RHCOS and avoids some CVE alerts ART was running into on some builds. I will go ahead and label this to merge to unblock 4.16 builds, we can discuss Monday if necessary.

PR #222, opened 2024-10-15
create 4.19 config file

ashwindasr:
As part of OCP branching step. Copied the [4.18 config file](https://github.com/openshift/check-payload/blob/main/dist/releases/4.18/config.toml)

No changes were added

### DIFF
--- a/dist/releases/4.19/config.toml
+++ b/dist/releases/4.19/config.toml
@@ -1,0 +1,433 @@
+[[payload.ose-network-interface-bond-cni-container.ignore]]
+error = "ErrLibcryptoSoMissing"
+files = ["/bondcni/bond", "/bondcni/rhel9/bond"]
+
+[[payload.ose-network-tools-container.ignore]]
+error = "ErrLibcryptoSoMissing"
+files = ["/usr/bin/ovnkube-trace"]
+
+[[payload.ose-machine-config-operator-container.ignore]]
+error = "ErrLibcryptoSoMissing"
+files = ["/usr/bin/machine-config-daemon.rhel9"]
+
+[[payload.ose-node-container.ignore]]
+error = "ErrLibcryptoSoMissing"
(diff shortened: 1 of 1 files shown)


dist/releases/4.19/config.toml
```diff
@@ -0,0 +1,433 @@
...
+  "/go/src/github.com/openshift/installer/terraform/bin/linux_arm64/terraform-provider-ovirt",
+  "/go/src/github.com/openshift/installer/terraform/bin/linux_arm64/terraform-provider-time",
+  "/go/src/github.com/openshift/installer/terraform/bin/linux_arm64/terraform-provider-vsphere",
+  "/go/src/github.com/openshift/installer/terraform/bin/linux_arm64/terraform-provider-vsphereprivate",
+]
+
+[[payload.ose-installer-terraform-providers-container.ignore]]
+error = "ErrGoNoCgoInit"
+files = [
+  "/go/src/github.com/openshift/installer/terraform/bin/linux_amd64/terraform",
```
mrunalp, line 187: We could benefit from expressing this whole list as `"/go/src/github.com/openshift/installer/terraform/bin/linux_amd64/terraform*"`

ashwindasr: @mrunalp we would want to do that for other files as well right? I can raise a follow up PR, since the changes do not relate to the branching task here, if that works

rhmdnd: While the existing implementation isn't very dry, it is explicit. Is this list of files everything under `/bin/linux_amd64/terraform*`? Or are there some files under here that aren't subject to ErrGoNoCgoInit?

rhmdnd (approved):
/lgtm and we can address the wildcard usage in a separate PR.

dist/releases/4.19/config.toml
ashwindasr: @rhmdnd I agree with you, if there are files that should not be subject to the exclude, then we should not use the wildcard. Not sure how we can verify that though

PR #225, opened 2024-10-22
fixes for check-payload for 4.15

rphillips:
The symbol import changed in go 1.20.12 from:
 `vendor/github.com/golang-fips/**openssl-fips**/openssl._Cfunc__goboringcrypto_DLOPEN_OPENSSL` to `vendor/github.com/golang-fips/**openssl**/openssl._Cfunc__goboringcrypto_DLOPEN_OPENSSL`

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -93,6 +93,10 @@ files = [
 error = "ErrNotDynLinked"
 files = ["/usr/libexec/podman/catatonit"]
 
+[[rpm.podman.ignore]]
+error = "ErrGoMissingSymbols"
+files = ["/usr/libexec/podman/rootlessport"]
+
 [[rpm.podman-catatonit.ignore]]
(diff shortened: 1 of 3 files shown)


joepvd:
Think config changes should also apply to 4.14, it has the same golang as 4.15. 

Symbols in 1.21 did not change. Change is in patch version of 1.20. These were recently tagged in and are related to this change: `golang-1.20.12-10.el9_3`, `golang-1.20.12-14.module+el8.9.0+22364+068108d4`.

sdodson:
/lgtm
I think the symbol change is accurate, only question I have is if the changes to add rhcos /unpack will be needed in 4.14 too. Given 4.14 isn't shipping this week going to go ahead and get this in, we can address additional changes in 4.14 in follow up, if necessary.

PR #223, opened 2024-10-16
OCPBUGS-43103: skip ErrNotDynLinked for catatonit in ansible-operator

chiragkyal:
Skip `executable is not dynamically linked` for `/usr/libexec/catatonit/catatonit` in [ansible-operator](https://github.com/openshift/ansible-operator-plugins/blob/d31f0a93e2523a67b5589c82884cc632331876d5/openshift/Dockerfile#L57)

`catatonit` has to be a static binary, and as per [Slack discussion](https://redhat-internal.slack.com/archives/CB95J6R4N/p1729083693856789), it does not perform any cryptographic operations.

https://pkgs.devel.redhat.com/cgit/rpms/catatonit/tree/catatonit.spec?h=rhel-9-main#n36

- Added fix for 4.17+
- Part of: https://issues.redhat.com/browse/OCPBUGS-43103

PR #242, opened 2025-01-22
[OSSM-8688] Ignore catatonit for kiali operator

mkralik3:
Fips scanner reports
```
Images with failures: ===================== image: registry.redhat.io/openshift-service-mesh/kiali-rhel9-operator@sha256:25a12f12294834a58f26792989432ca246cde43eb055ccda74c8a5c8af119ebe build: openshift-istio-kiali-operator-container-2.4.1-1 check-payload failures: * /usr/libexec/catatonit/catatonit: executable is not dynamically linked
```
for our kiali operator image. 

We use `ose-ansible-rhel9-operator:v4.17.0-202412170235` as a base image for which it was resolved by https://github.com/openshift/check-payload/pull/223 .

I have added kiali operator to ignore list as well.

### DIFF
--- a/dist/releases/4.16/config.toml
+++ b/dist/releases/4.16/config.toml
@@ -400,6 +400,10 @@ files = [
 error = "ErrNotDynLinked"
 files = ["/usr/libexec/catatonit/catatonit"]
 
+[[payload.openshift-istio-kiali-operator-container.ignore]]
+error = "ErrNotDynLinked"
+files = ["/usr/libexec/catatonit/catatonit"]
+
 [[payload.ose-cloud-credential-operator-container.ignore]]
 error = "ErrLibcryptoSoMissing"
 files = ["/usr/bin/ccoctl", "/usr/bin/ccoctl.rhel8"]
--- a/dist/releases/4.17/config.toml
+++ b/dist/releases/4.17/config.toml
@@ -388,6 +388,10 @@ files = [
 error = "ErrNotDynLinked"
 files = ["/usr/libexec/catatonit/catatonit"]
 
+[[payload.openshift-istio-kiali-operator-container.ignore]]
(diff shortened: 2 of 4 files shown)


config.toml
```diff
@@ -164,3 +164,7 @@ dirs = ["/assets/downloads/cli"]
 [[payload.rhacs-main-container.ignore]]
 error = "ErrGoMissingTag"
 dirs = ["/assets/downloads/cli"]
+
+[[payload.openshift-istio-kiali-operator-container.ignore]]
```
rhmdnd, line 168: Is this container image and binary present in all versions of OpenShift?

If not, we may want to consider adding it to the version specific configs in https://github.com/openshift/check-payload/tree/main/dist/releases - which is the approach used in https://github.com/openshift/check-payload/pull/223/files.

mkralik3: @rhmdnd  updated

rhmdnd: Ok - I appreciate the clarification. Thanks for breaking it out into the versioned configs.

dist/releases/4.16/config.toml
```diff
@@ -400,6 +400,10 @@ files = [
 error = "ErrNotDynLinked"
 files = ["/usr/libexec/catatonit/catatonit"]
 
+[[payload.openshift-istio-kiali-operator-container.ignore]]
+error = "ErrNotDynLinked"
+files = ["/usr/libexec/catatonit/catatonit"]
```
rhmdnd, line 405: Note for other reviews and historical reference. The catatonic binary is built statically, hence the exception here.

config.toml
ferhoyos: Why didn't we include this config for OCP 4.14 and OCP 4.15? AFAIK these OCP versions are to be supported by the Kiali operator.

PR #246, opened 2025-02-20
added missing ignore and extand to all versions

tal-hason:
after validating the last update we still saw missing configurations, so we added them and extended to all versions just to be on the safe side, cause we noticed in v4.12 didn't catch the ignores


rhmdnd (commented):
This change is adding an incredible amount of exclusions.

Are all of these needed? Do you have more context around what specifically is failing? What component, container image, or payload are you running this against to get this many failure?

tal-hason:
> This change is adding an incredible amount of exclusions.
> 
> Are all of these needed? Do you have more context around what specifically is failing? What component, container image, or payload are you running this against to get this many failure?

@rhmdnd , we noticed that the fips check breaks once it founds unignored fips related files.

So we ran the scan locally each bundle over and over again and found those, new once.

@Axuba if you have anything else to add?

tal-hason:
merged in https://github.com/openshift/check-payload/pull/248

PR #248, opened 2025-02-28
Add FIPS exclusions to CNV

Axuba:
I've added (and tested locally) the exclusions we were missing on CNV. Those exclusions are on binaries that are on test containers that are not shipped to customers, or go binaries used by those same test containers.

### DIFF
--- a/dist/releases/4.17/config.toml
+++ b/dist/releases/4.17/config.toml
@@ -449,3 +449,47 @@ files = ["/usr/bin/hcp-no-cgo", "/usr/bin/hypershift-no-cgo"]
 [[payload.multicluster-engine-hypershift-operator-container.ignore]]
 error = "ErrGoNotCgoEnabled"
 files = ["/usr/bin/hcp-no-cgo", "/usr/bin/hypershift-no-cgo"]
+
+[[payload.virt-cdi-importer-rhel9-container.ignore]]
+error = "ErrGoNotCgoEnabled"
+files = ["/usr/bin/cdi-containerimage-server"]
+
+[[payload.virt-cdi-importer-rhel9-container.ignore]]
+error = "ErrGoMissingSymbols"
+files = ["/usr/bin/cdi-containerimage-server"]
+
+[[payload.virt-cdi-importer-rhel9-container.ignore]]
+error = "ErrNotDynLinked"
+files = ["/usr/bin/cdi-containerimage-server"]
+
+[[payload.virt-cdi-importer-rhel9-container.ignore]]
+error = "ErrLibcryptoMissing"
+files = ["/usr/bin/cdi-containerimage-server"]
+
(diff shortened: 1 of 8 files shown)


Axuba:
Hello @rhmdnd , could you please review this PR, and ignore https://github.com/openshift/check-payload/pull/246 ?

dist/releases/4.17/config.toml
```diff
@@ -448,4 +448,48 @@ files = ["/usr/bin/hcp-no-cgo", "/usr/bin/hypershift-no-cgo"]
...
+
+[[payload.virt-cdi-importer-rhel9-container.ignore]]
+error = "ErrNotDynLinked"
+files = ["/usr/bin/cdi-containerimage-server"]
+
+[[payload.virt-cdi-importer-rhel9-container.ignore]]
+error = "ErrLibcryptoMissing"
+files = ["/usr/bin/cdi-containerimage-server"]
+
+[[payload.kubevirt-tekton-tasks-create-datavolume-rhel9-container.ignore]]
```
rhmdnd, line 469: Just curious, how come there are missing for 4.15 and 4.16 releases? Are those branches/releases not moved over to Konflux, yet?

rhmdnd (approved):
/lgtm

dist/releases/4.19/config.toml
```diff
@@ -473,49 +472,60 @@ files = ["/usr/bin/container-disk"]
 
 [[payload.virt-cdi-importer-rhel9-container.ignore]]
 error = "ErrGoNotCgoEnabled"
-files = [
-"/usr/bin/cdi-containerimage-server"
-]
+files = ["/usr/bin/cdi-containerimage-server"]
+
+[[payload.virt-cdi-importer-rhel9-container.ignore]]
```
dankenigsberg, line 477: @dominikholler  @aglitke can you add a one-line comment explaining why each of these errors is OK for each of the containers?

dist/releases/4.17/config.toml
tal-hason: @rhmdnd , we are still not migrated to konflux, this is on the roadmap for this year

dankenigsberg:
> I've added (and tested locally) the exclusions we were missing on CNV. Those exclusions are on binaries that are on test containers that are not shipped to customers, or go binaries used by those same test containers.

@Axuba @tal-hason thanks for helping CNV pass the tests.

I may miss something big here, [virt-handler](https://github.com/openshift/check-payload/pull/248/files#diff-a23c252d37492604ada02036c9670822867dd14766576cf690b171253559443aR41) is not a test container. We have a good reason why `/usr/bin/container-disk` is statically-linked, but it need to be specified in `config.toml` or at least in the message of the commit introducing it. Otherwise we risk blanket ignoring of these important tests.

PR #249, opened 2025-03-03
Add virt-handler and virt-launcher ignores that weren't reported by CVP

Axuba:
I made last PR https://github.com/openshift/check-payload/pull/248 based on the result of the FIPS check on CVP. However, CVP was using for versions v4.14 and up the 4.19 configuration file. On konflux this is not like this anymore, konflux uses the respective version (for example v4.14 uses the 4.14 configuration file). A couple of ignores were missing because of that. I just replicated them on the missing versions.


rhmdnd (commented):
Changes look fine - looks like the base branch needs to be updated, though.

PR #250, opened 2025-03-12
Turn off usage on scan failures

dhaiducek:
- Turn off usage output on scan failures
- Prefer `Run()` when no errors are returned
- Put config output behind verbosity flag

### DIFF
--- a/internal/types/types_config.go
+++ b/internal/types/types_config.go
@@ -9,7 +9,7 @@ import (
 )
 
 func (c *Config) Log() {
-	klog.Infof("using config %+v", c)
+	klog.V(1).Infof("using config %+v", c)
 }
 
 // isMatch tells if path equals to one of the entries.
--- a/main.go
+++ b/main.go
@@ -72,9 +72,8 @@ func main() {
 	versionCmd := &cobra.Command{
 		Use:   "version",
 		Short: "Print the version",
-		RunE: func(_ *cobra.Command, _ []string) error {
+		Run: func(_ *cobra.Command, _ []string) {
 			fmt.Println(Commit)
-			return nil
 		},
 	}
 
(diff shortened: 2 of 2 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/openshift/check-payload
 
-go 1.19
+go 1.21
```
kolyshkin, line 3: 1. The commit title is somewhat deceiving -- this is more than `go mod tidy`, you are bumping the minimally supported Go version. Please say so.

2. If you're bumping it, you need to remove `internal/golang/elf_go120.go` as it is no longer needed. Maybe there's something more.

3. If you are to bump it, why not bump to Go 1.23 right away (Go 1.22 is no longer supported).

4. It makes sense to also modify CI (`.github/workflows/validate.yml`) to test on supported Go versions (1.23.x, 1.24.x).

5. I did the same change locally, but it did not result in the same change to imdario/mergo that your commit did. Hmm

kolyshkin (changes requested):
LGTM except for the `go mod tidy` commit (see above).

go.mod
kolyshkin: 6. Go version in Dockerfile might need to be updated, too.

dhaiducek: My fault. Those are entirely valid points and this commit doesn't align with this PR--I'll move it to a separate PR for review.

The reproducer for this commit is:
- Remove the `replace` statement (aside from being cleaner, its removal should enable a `go install` of the binary)
- `go get github.com/imdario/mergo@v0.3.16`
- `go mod tidy`

I honestly wasn't expecting the Go bump with `go mod tidy`, but it's definitely inappropriate for this PR.

dhaiducek: Opened up a new PR for the Go bump, including items listed (and other instances of Go versions I found in devcontainer and `go:build` statement in the `elf` file):
- #252

kolyshkin (approved):
LGTM

PR #251, opened 2025-03-12
Fix `Dockerfile.upstream`

dhaiducek:
- Realignment with `Dockerfile`
- Add missing `wget`

### DIFF
--- a/Dockerfile.upstream
+++ b/Dockerfile.upstream
@@ -11,14 +11,14 @@ FROM fedora:38
 ARG OC_VERSION=latest
 ARG UMOCI_VERSION=latest
 
-RUN curl -LJo "openshift-client-linux-${OC_VERSION}.tar.gz" "https://mirror.openshift.com/pub/openshift-v4/amd64/clients/ocp/${OC_VERSION}/openshift-client-linux.tar.gz" \
+RUN dnf -y update && dnf install -y wget binutils go file runc podman jq skopeo && dnf clean all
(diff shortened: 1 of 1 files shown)


Dockerfile.upstream
```diff
@@ -11,14 +11,14 @@ FROM fedora:38
 ARG OC_VERSION=latest
 ARG UMOCI_VERSION=latest
 
-RUN curl -LJo "openshift-client-linux-${OC_VERSION}.tar.gz" "https://mirror.openshift.com/pub/openshift-v4/amd64/clients/ocp/${OC_VERSION}/openshift-client-linux.tar.gz" \
+RUN dnf -y update && dnf install -y wget binutils go file runc podman jq skopeo && dnf clean all
+RUN wget -O "openshift-client-linux-${OC_VERSION}.tar.gz" "https://mirror.openshift.com/pub/openshift-v4/amd64/clients/ocp/${OC_VERSION}/openshift-client-linux.tar.gz" \
```
kolyshkin, line 15: What is the motivation behind the curl -> wget change? Might make sense to add it to commit description.

dhaiducek: It's a copy/paste from `Dockerfile`, which I was taking as a kind of source of truth for the fix/sync:
https://github.com/openshift/check-payload/blob/09a33418ded3f9217f8e511076ded4613f04ec59/Dockerfile#L15-L16

kolyshkin: Makes sense

kolyshkin (approved):
LGTM, thanks!

dhaiducek:
Looks like this is blocked until this Prow PR is merged:
- https://github.com/openshift/release/pull/62752

PR #252, opened 2025-03-12
Upgrade to Go1.23 and OCP 4.18

dhaiducek:
- Upgrade to Go v1.23
- Remove `replace` from `go.mod`, which opens the way for `go install`
- Remove GH workflow files since Actions are disabled for the org
- Remove `elf` go1.20 workaround
- Upgrade to OCP 4.18

Split from/Addresses comments from:
- https://github.com/openshift/check-payload/pull/250#discussion_r1990739413


.github/workflows/validate.yml
```diff
@@ -7,11 +7,14 @@ on:
       - main
       - release-*
   pull_request:
-env:
-  GO_VERSION: 1.22.x
 
 jobs:
   lint:
```
dhaiducek, line 12: I just realized this isn't running in this PR, but it is running in Prow. 🤔

dhaiducek:
Looks like this is going to be blocked unless the Prow images are also upgraded. I've opened that PR here (and also took the liberty to bumped the OCP version to 4.18, but I can revert that if it's undesirable):
- https://github.com/openshift/release/pull/62718

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.21-openshift-4.16 AS builder
+FROM registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.23-openshift-4.16 AS builder
```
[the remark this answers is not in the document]
dhaiducek: Looks like the `openshift/release` PR was approved, so I've added a commit to upgrade here also.

internal/golang/elf.go
```diff
@@ -1,4 +1,4 @@
-//go:build go1.21
+//go:build go1.23
```
kolyshkin, line 1: This is not necessary to change Go version here. Whatever's in the file works since go1.21, and go1.23 sets all the previous versions tags.

kolyshkin:
LGTM overall, just a couple of nits.

.github/workflows/validate.yml
dhaiducek: Well. 😕 Do you want me to go ahead and delete the file in this PR, then?

kolyshkin: I guess so, yeah (unless you know a way to re-enable GHA:)

dhaiducek: Haha, unfortunately I don't have such powers 😆

dhaiducek:
Looks like this is blocked by my Prow missteps, fixed in this PR:
- https://github.com/openshift/release/pull/62752

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/openshift/check-payload
 
-go 1.19
+go 1.23.0
```
rphillips, line 3: should be 1.23

dhaiducek: Huh. I thought 1.23 had to have a z-stream specified. 😕 I'll update it.

kolyshkin: In fact it should be 1.23.0 (they have changed the notation at around go 1.20 or 1.21).

rphillips: I think 1.23 is correct based on the [docs](https://go.dev/doc/modules/gomod-ref#go).

dhaiducek: It's weird--I could have sworn a different upgrade to 1.23 was upset with me that there wasn't a z-stream, but I tried it here and it all seems to be fine? I'll give it a shot at least.

rphillips:
We need to bump the internal/golang/ package to include a 1.23 Pie module.

rphillips:
Perhaps:

```
diff --git i/internal/golang/elf.go w/internal/golang/elf.go
index 580ceff..328ff11 100644
--- i/internal/golang/elf.go
+++ w/internal/golang/elf.go
@@ -1,4 +1,4 @@
-//go:build go1.21
+//go:build go1.21 || go1.22 || go1.23 || go1.24

 package golang


```

rphillips:
Darn... Perhaps we do have to upgrade to golangci-lint:v1.63.4.

```
❯ docker run --rm --entrypoint /bin/sh -it quay.io/openshiftci/golangci-lint:v1.63.4
# go version
go version go1.23.4 linux/amd64
```

dhaiducek:
Back to Prow! PR opened:
- https://github.com/openshift/release/pull/62821

kolyshkin:
> Perhaps:
> 
> ```diff
> diff --git i/internal/golang/elf.go w/internal/golang/elf.go
> index 580ceff..328ff11 100644
> --- i/internal/golang/elf.go
> +++ w/internal/golang/elf.go
> @@ -1,4 +1,4 @@
> -//go:build go1.21
> +//go:build go1.21 || go1.22 || go1.23 || go1.24
> 
>  package golang
> ```

But why? This is _totally_ not needed. Each Go 1.x release adds go1.(x-1), go1.(x-2) etc. build tags.

internal/golang/elf.go
```diff
@@ -1,4 +1,4 @@
-//go:build go1.21
+//go:build go1.21 || go1.22 || go1.23 || go1.24
```
kolyshkin, line 1: Please remove, this is not needed.

kolyshkin:
The errors in CI probably means that old Go version is used.

kolyshkin (commented):
Post-merge LGTM, thanks!

go.mod
kolyshkin: So, `1.23` means *any* 1.23 release, including `rc` etc. `1.23.0` means the same thing but without rcs, betas etc. The difference is subtle and probably doesn't matter in practice, but I saw that recently go tools start to add the .z suffix.

dhaiducek:
Nice! Sure thing! `go install` works, too! 🎉 

```
go install github.com/openshift/check-payload@main
```

...though the version metadata doesn't come through in `check-payload version`, and for some reason the `latest` tag is also the latest commit 🤔 It looks like it's because the tags don't have a leading `v`, because using `@0.3.4` returns the expected "contains replace directives". (Versions in `pkg.go.dev`: https://pkg.go.dev/github.com/openshift/check-payload?tab=versions)

```
$ curl https://proxy.golang.org/github.com/openshift/check-payload/@latest

{
  "Version":"v0.0.0-20250314174204-37cd141d04b8",
  "Time":"2025-03-14T17:42:04Z",
  "Origin":{
    "VCS":"git","URL":"https://github.com/openshift/check-payload","Hash":"37cd141d04b8cf49cee0e6ee29b0c8f49c1fff98"
  }
}
```

PR #269, opened 2025-06-16
Update OCP virt validation exception binary path

Axuba:
Follow up of https://github.com/openshift/check-payload/pull/265

The binary path and name changed, so we are requesting to update the exception to the new name and path. Nothing else changed.


rhmdnd (approved):
Looks good. Was the change primarily project structure or build refactoring?

PR #271, opened 2025-06-19
volsync-container exception for diskrsync binary (for ocp 4.14)

tesshuflower:
- Back-porting exception that was only done from ocp 4.15 and up since that was the oldest CVP would use when running check-payload

- Now that we are building File-based catalogs for OCP versions in konflux, it appears the check-payload is run using the OCP 4.14 config for the ocp 4.14 FBC and so we get the failure again for `diskrsync`.

- Previous PR for reference: https://github.com/openshift/check-payload/pull/177


rhmdnd (approved):
Looks consistent with the 4.15+ changes.

/lgtm

Issue #272, opened 2025-06-20
Check-payload with specific OCP 4.16 configuration has error on some images


tesshuflower:
I have an image here which is failing the check-payload scan when run with `-V 4.16`:

```
quay.io/redhat-user-workloads/volsync-tenant/volsync-0-13@sha256:4318a1af43e17bfa865594eae08d63e2aae705889ed35c39c9eece14f67d443e
```

It appears this is because only 4.16 has `certified_distributions` set in the config.toml:  https://github.com/openshift/check-payload/blob/main/dist/releases/4.16/config.toml#L1-L5

And the image above is using this as a base image:  [registry.redhat.io/ubi9/ubi-minimal:latest](http://registry.redhat.io/ubi9/ubi-minimal:latest)

The image in question has `/etc/redhat-release` that looks like this:  
```
Red Hat Enterprise Linux release 9.6 (Plow)
```

It seems that just adding the distro above to the list would fix it, but there's a second strange issue which is that I have other images built using the same base image and they pass the check-payload scan with `-V 4.16`.

Here's an example image which does **not** fail the scan:  
```
quay.io/redhat-user-workloads/crt-redhat-acm-tenant/cluster-backup-operator-acm-214@sha256:148faf995f4090cfaf4baaadae728fc10919ee8863154dfd2e08bf214cb59e5e
```

This also has the same thing for /etc/redhat-release, so not sure why this one passes.

tesshuflower:
This issue has come up again, as the latest ubi9-minimal is now:

`Red Hat Enterprise Linux release 9.7 (Plow)`

Do we need this specific list of certified distros for OCP 4.16 only?  It seems other OCPs don't have this check

tesshuflower:
It actually looks like RHEL 9.7 is added to the global config.toml now in the latest code in `main`: https://github.com/openshift/check-payload/blob/main/config.toml#L13

And there is no longer any special list only for OCP 4.16.

Is there a chance we can get a version tagged with this update? We need this for our konflux FBC builds to pass

PR #273, opened 2025-06-26
add RHEL 9.6 to certified_distributions for OCP 4.16

tesshuflower:
Adds RHEL 9.6 to the list of certified distributions for OCP 4.16.

Related issue: https://github.com/openshift/check-payload/issues/272

Note this does not resolve everything in the above issue, but will unblock failing scans on OCP 4.16 that use a RHEL 9.6 base image.

Remaining questions/problems from the above issue:
- Why does only OCP 4.16 have this list of certified distributions in the config.toml?
- Why do many images built with the same RHEL 9.6 base image pass the scan without this fix?

### DIFF
--- a/dist/releases/4.16/config.toml
+++ b/dist/releases/4.16/config.toml
@@ -1,6 +1,7 @@
 certified_distributions = [
   "Red Hat Enterprise Linux release 9.2 (Plow)",
   "Red Hat Enterprise Linux release 9.4 (Plow)",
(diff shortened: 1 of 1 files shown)


rhmdnd:
Should be good since OpenSSL in 9.6 is covered under the same validation as 9.4 and 9.2.

https://access.redhat.com/compliance/fips
https://csrc.nist.gov/projects/cryptographic-module-validation-program/certificate/4857

PR #277, opened 2025-07-23
fix: skipped os scan errors for certified distributions

smith-xyz:
Looking into #272 there appeared to be a discrepancy between different images being validated for FIPS certified OS. This led to addressing a few issues:

1. Only components that had an entry in the config.toml would be validated on the FIPS OS, fixed this logic error.
2. Only version 4.16 config was checking certified distributions, moved this to global so all versions are checked.
3. After doing so, noticed FIPS certified OSs missing, added those to the list.
4. Noticed the validation of the os would not handle symlinked red-release files, added logic to handle this scenario.
5. Added RHCOS versions to certified distribution list for container images that used symlinked /etc/redhat-release -> /usr/lib/system-release

Added tests to demonstrate what I believe to be the expected behavior for this validation

Tested a few payloads to ensure they were passing:
```bash
./check-payload scan payload -V 4.20 --url quay.io/openshift-release-dev/ocp-release:4.20.0-ec.2-multi-x86_64 --output-file 4.20.0-ec.2-multi-x86_64_report.txt

./check-payload scan payload -V 4.16 --url quay.io/openshift-release-dev/ocp-release:4.16.42-x86_64 --output-file 4.16.42-x86_64_report.txt

./check-payload scan payload -V 4.12 --url quay.io/openshift-release-dev/ocp-release:4.12.77-x86_64 --output-file 4.12.77-x86_64_report.txt
```

### DIFF
--- a/internal/scan/scan.go
+++ b/internal/scan/scan.go
@@ -377,9 +377,7 @@ func walkDirScan(ctx context.Context, cfg *types.Config, tag *v1.TagReference, c
 		// operating systems that haven't been certified, yet.
 		osInfo := validations.ValidateOS(cfg, mountPath)
 		osScanResult := types.NewScanResult().SetOS(osInfo).SetComponent(component).SetTag(tag)
-		if osScanResult.Error != nil {
-			results.Append(osScanResult)
-		}
+		results.Append(osScanResult)
 	}
 
 	// does the image contain openssl


smith-xyz:
removed RHCOS from cert distribution list to avoid confusion

ran against nightly 4.16-4.20, all is good except rhel-coreos on 4.16-4.18. Not sure why, but we can't extract component information and for some reason this was the one image built on RHCOS.

rhmdnd:
Looks good to me. I think it would be good to have an understanding of how/when we should be using the RHCOS image though. From a payload perspective, we should be primarily using ubi images, right? Maybe someone from the RHCOS/Fedora CoreOS team can help us understand that a little better (@jlebon perhaps?).

rhmdnd:
/lgtm

Built and ran against 4.16 payloads without issues.

PR #280, opened 2025-08-12
ART-13785: create 4.21 config file

mbiarnes:


mbiarnes:
As part of OCP branching step. Copied the [4.20 config file](https://github.com/openshift/check-payload/blob/main/dist/releases/4.20/config.toml)

No changes were added

PR #282, opened 2025-09-19
The MCE hypershift container needs changes to the exceptions in place

gparvin:
The FIPS scans are failing in our konflux builds due to problems with the hypershift "extra" binaries.  The current exceptions for these binaries is not correctly preventing the scanner from erroring on these binaries. Tested with the latest image from our MCE 2.18.3 release

Refs:
 - https://issues.redhat.com/browse/ACM-24281

### DIFF
--- a/dist/releases/4.15/config.toml
+++ b/dist/releases/4.15/config.toml
@@ -178,7 +178,23 @@ files = [
 
 [[payload.multicluster-engine-hypershift-operator-container.ignore]]
 error = "ErrGoNotCgoEnabled"
-files = ["/usr/bin/hcp-no-cgo", "/usr/bin/hypershift-no-cgo"]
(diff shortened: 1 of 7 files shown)


rhmdnd:
> @rhmdnd @gparvin @rphillips - I believe the exception just adds "/usr/bin/hcp" and is used for QE testing.
> 
> I'm wondering if it is worth adding some comments around these exceptions to make clear why the exceptions are being added, noting how the inclusion of said binaries impacts customers directly/indirectly, does it ever get filtered out later in some other process, etc.

+1

Yeah, comments would help at least a little bit. If the `hcp` binary is only used for testing purposes, does it have to be in the payload container? Can it be distributed to people who need to test with it in a different way such that it's not exposed to users?

kolyshkin:
@gparvin this needs to be rebased, plus https://github.com/openshift/check-payload/pull/282#issuecomment-3745033493 needs an answer

gparvin:
Closing this pr in favor of https://github.com/openshift/check-payload/pull/339

PR #295, opened 2025-11-17
ART-14216: create 4.22 config file

mbiarnes:


ashwindasr:
@mbiarnes to confirm, this is a copy of 4.21 file right, with no changes?

PR #302, opened 2025-12-05
Ignore rhel-coreos-10 image ErrOSNotCertified for now

sdodson:
4.21 and 4.22 will have RHCOS 10 images but these are not GA therefore not beholden to FIPS constraints. We will ensure that documentation when these features are made TechPreview specifically mentions that they should not be used in FIPS regulated environments.


rhmdnd (commented):
/lgtm

Adding @smith-xyz and @richardsonnick for awareness.

PR #309, opened 2026-01-16
chore: add rhel 10 certified distros

smith-xyz:


smith-xyz:
going to close this for now, I don't think certification is complete yet. Will triple check but nothing on nist is listed.

PR #316, opened 2026-03-19
feat(validations): add module-based FIPS validation

### LINKED ISSUE #315: Certified distribution check: module-specific validation instead of allowlist
The certified distribution check is an allowlist of OS release strings (from /etc/redhat-release). It does not verify that the image actually ships a CMVP validated crypto module; RHEL as an OS is not FIPS certified—certification applies to the cryptographic modules (e.g. OpenSSL FIPS provider, GnuTLS, NSS) that ship in the distribution. The tool only matches the release file against the list, so the list is a policy proxy and the implementation never validates that a certified module is present. A better implementation would tie the check to the modules components actually use: for example, if a component uses OpenSSL, the tool could enforce that the base meets the bar for that module (e.g. el9 >= 9.4, where the crypto team does not ship an unvalidated OpenSSL module) rather than a global allowlist of release strings. The open question is how to evolve the check in this tool toward that—whether to validate presence/version of known certified modules, to drive checks by component and the modules each uses, or to keep the allowlist but document and maintain it in a more principled way.

Proposed direction: drive the certified-base check by the crypto modules the scan actually finds in use (the tool already detects OpenSSL via libcrypto linkage and validates it at image level). For each such module, apply a version rule for the base OS instead of a single global allowlist—e.g. when OpenSSL is in use, require RHEL 9 base >= 9.4 and, once documented, RHEL 10 when the OpenSSL OE update is in place. Components that do not use a tracked module could remain subject to the current allowlist or be excluded from the OS check. That would align the implementation with how certification actually works (per module) and reduce drift as new RHEL minors ship. It would also be nice to automate adding or updating those rules (e.g. from Red Hat's FIPS/CMVP documentation) so they stay in sync as new certified modules or releases ship.

```mermaid
flowchart LR
  A[Scan] --> B[Base OS]
  B --> C[Modules in use]
  C --> D[Per-module rules]
  D --> E{Pass?}
  E --> F[OK]
  E --> G[Fail]
```

smith-xyz:
This PR adjusts the OS validation to be accompanied by a fips crypto module validation. This will provide us with:
- a more accurate check for a fips approved module based on config
- the flexibility to add future modules like go crypto when FIPS certification is provided for that module

Resolves #315


smith-xyz:
This abstraction is for #332 as well

PR #317, opened 2026-04-07
fix(config): remove rhel 9.8 beta from 4.22 certified distributions

smith-xyz:
removes 9.8 beta from certified os list


sdodson:
/hold
Need to wait on new builds that drop beta term

sdodson:
@smith-xyz We could remove all references to Beta at this point because both 9.8 and 10.2 content we have today are no longer Beta. Do you mind updating to get all of those occurences in dist/releases/ ?

PR #326, opened 2026-04-24
Exempt vllm-gaudi /usr/bin/hl-smi

RH-steve-grubb:
hl-smi is the main tool for monitoring a gaudi hardware accelerator. It cannot be removed. It is built by Intel so there's nothing we can do to solve the problem. It is likely static so that it can run on Red Hat or Ubuntu without worrying is library versions are compatible.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -158,6 +158,10 @@ files = [
   "/usr/libexec/valgrind/none-amd64-linux",
 ]
 
+[[payload.odh-vllm-gaudi-rhel9.ignore]]
+error = "ErrNotDynLinked"
+files = ["/usr/bin/hl-smi"]
+
 [[payload.openshift-enterprise-pod-container.ignore]]
 error = "ErrNotDynLinked"
 files = ["/usr/bin/pod"]
--- a/config.toml
+++ b/config.toml
@@ -158,7 +158,8 @@ files = [
   "/usr/libexec/valgrind/none-amd64-linux",
 ]
 
-[[payload.odh-vllm-gaudi-rhel9.ignore]]
+[[rpm.habanalabs-firmware-tools.ignore]]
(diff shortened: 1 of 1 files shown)


smith-xyz:
I get Intel ships hl-smi static for portability — that's their call. But just to clarify: why does it need to be in the odh-vllm-gaudi-rhel9 payload image rather than available on the host or in a separate tools image?

Does the vLLM runtime invoke hl-smi? If it's only used for operator/admin monitoring, does it belong at the node level or in a dedicated diagnostics image, not baked into every inference container?

RH-steve-grubb:
This is needed for [RHOAIENG-58639](https://redhat.atlassian.net/browse/RHOAIENG-58639)
This is only needed for gaudi systems - so it belongs in the gaudi image. You can think of hl-smi as their equivalent of nvidia-smi. nvidia-smi is in every vllm-cuda image. Because you may need to login and see how much memory is occupied in the accelerator. Putting it in another image just moves the problem. It is still a problem.

The main thing to know is that hl-smi does not do any cryptography. And even if it did, FIPS only applies to user data. This is a system tool strictly for troubleshooting the hardware.

https://docs.habana.ai/en/latest/Management_and_Monitoring/Embedded_System_Tools_Guide/System_Management_Interface_Tool.html

ckhordiasma:
@RH-steve-grubb can you update to the following? this will prevent us from having to submit an exception for every RHOAI release, as the payload method's name needs to match `com.redhat.component` label, which is version specific for us

```
[[rpm.habanalabs-firmware-tools.ignore]]
# hl-smi: Intel Gaudi hardware monitor, static by design, no crypto.
error = "ErrNotDynLinked"
files = ["/usr/bin/hl-smi"]
```

RH-steve-grubb:
@ckhordiasma The PR has been updated as specified.

PR #327, opened 2026-04-28
RHOAIENG-58626: Ignore ErrNotDynLinked for pandoc, py-spy, and rg

ysok:
[RHOAIENG-58626](https://redhat.atlassian.net/browse/RHOAIENG-58626): Ignore ErrNotDynLinked for pandoc, py-spy, and rg

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -214,3 +214,144 @@ files = ["/usr/local/bin/mta-cli"]
 [[payload.mta-cli-container.ignore]]
 error = "ErrNotDynLinked"
 files = ["/usr/local/bin/mta-cli"]
+
+# -----------------------------------------------------------------------------
+# Open Data Hub (ODH) / RHOAI payloads: odh-workbench-* and odh-pipeline-runtime-*
+#
+# odh-workbench-* images are interactive JupyterLab and code-server workbenches.
+# Pandoc and ripgrep (via code-server) cover notebook PDF export and IDE search.
+#
+# py-spy is a Python sampling profiler brought in with the datascience/ML dependency
+# graph so users can debug performance from the workbench. It does not use
+# cryptography. Upstream distributes it as a statically linked executable, so
+# check-payload raises ErrNotDynLinked; we waive that here because the finding reflects
+# static linking, not an in-image cryptographic module or TLS endpoint.
+#
+# odh-pipeline-runtime-* images are non-interactive Elyra/KFP-style job runtimes; see the
+# comment block immediately before those entries for py-spy on pipeline images.
+#
+# ErrNotDynLinked ignores cite RHOAIENG-58626 (umbrella epic for multiple check-payload
+# FIPS findings across RHOAI). py-spy waivers: track RHOAIENG-58916. Work is broken
+# down by failing component/image in Jira. Pandoc rebuild/remediation: AIPCC-7795.
+# -----------------------------------------------------------------------------
+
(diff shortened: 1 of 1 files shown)


smith-xyz:
Please address the following:

- Scope these to specific payload components ([[payload.<notebook-component>.ignore]]), not [[ignore]]. That is global so we don't want to apply that equally to all components.
- Why are py-spy, pandoc, and rg in a production payload image? These are dev/convenience tools. If they're not required for the supported runtime, remove them from the image rather than ignoring the check.
- If they truly can't be removed, each needs its own justification with a tracking issue for upstream resolution — not a single blanket ignore block.
- "Transitive dependency" is a reason to fix the dep, not to ignore the check.

ckhordiasma:
@ysok I'll defer to your team for a better explanation.. but my understanding is that these notebooks images--there are several, which is why an blanket ignore exception was requested--are essentially to provide dev environment-like containers for customers, right?

ysok:
@smith-xyz @ckhordiasma I just made another push with separation of each image with the jira ticket. Is this good enough or anything else needed? 

Discussion at slack: https://redhat-internal.slack.com/archives/C07SBP17R7Z/p1776781639457179?thread_ts=1776778459.956129&cid=C07SBP17R7Z

smith-xyz:
@ysok much better than you!

A few things before I can `/hold cancel`:

- Add a context block above the ODH section explaining that these are interactive workbench/notebook images (customer-facing dev environments), not traditional production workloads. That justification currently only lives in Slack and will get lost.

- py-spy in pipeline runtime images — workbench images being dev environments makes sense for a profiler, but why do odh-pipeline-runtime-* images need py-spy? Pipeline runtimes aren't interactive. If it's a shared base image issue, note that in the comment.

- Single tracking issue — [RHOAIENG-58626](https://redhat.atlassian.net/browse/RHOAIENG-58626) covers py-spy, pandoc, and rg (three unrelated upstream tools). Is there a plan to track upstream resolution for each independently, or is the ODH team owning all three?

Once those are addressed I'm happy to approve.

ysok:
> @ysok much better than you!
> 
> A few things before I can `/hold cancel`:
> 
> * Add a context block above the ODH section explaining that these are interactive workbench/notebook images (customer-facing dev environments), not traditional production workloads. That justification currently only lives in Slack and will get lost.
> * py-spy in pipeline runtime images — workbench images being dev environments makes sense for a profiler, but why do odh-pipeline-runtime-* images need py-spy? Pipeline runtimes aren't interactive. If it's a shared base image issue, note that in the comment.
> * Single tracking issue — [RHOAIENG-58626](https://redhat.atlassian.net/browse/RHOAIENG-58626) covers py-spy, pandoc, and rg (three unrelated upstream tools). Is there a plan to track upstream resolution for each independently, or is the ODH team owning all three?
> 
> Once those are addressed I'm happy to approve.

Updated per your comment above.

Ticket: [RHOAIENG-58626](https://redhat.atlassian.net/browse/RHOAIENG-58626) covers py-spy, pandoc, and rg. Also we have each invidual tickets for py-spy RHOAIENG-58916, pandoc rebuild AIPCC-7795. For ripgrep, it is hard requirement for CodeServer.

smith-xyz:
/hold cancel
/lgtm
/approve


thank you - going to move this along to unblock you all. Just to note that I think we could improve this whole config setup to be a bit cleaner, maybe component based over global/ocp version based. But for another time.

PR #330, opened 2026-05-09
fix: check .gopclntab before .data.rel.ro for PIE binaries

### LINKED ISSUE #329: `ReadTable` fails on Go 1.26 PIE binaries that have a separate `.gopclntab` section
## Summary

`check-payload scan local` fails with "could not find magic number" on Go 1.26-built PIE binaries because `ReadTable` in `internal/golang/goscan.go` unconditionally redirects to `.data.rel.ro` when it sees `-buildmode=pie`. Go 1.26 changed the ELF layout for PIE builds: it now emits `.gopclntab` as a **separate section** instead of embedding the pclntab in `.data.rel.ro` (which was the Go 1.25 and earlier behavior).

This will affect **all OCP components** once the RHEL 9 Go toolchain moves to 1.26, not just the images where we first observed it.

## Root Cause

The section lookup logic in [`internal/golang/goscan.go`](https://github.com/openshift/check-payload/blob/main/internal/golang/goscan.go) overrides the section label when the binary is PIE:

```go
sectionLabel := ".gopclntab"
for _, bs := range bi.Settings {
    if bs.Key == "-buildmode" && bs.Value == "pie" {
        sectionLabel = ".data.rel.ro"   // <-- skips .gopclntab entirely
        break
    }
}
```

In Go <= 1.25, PIE binaries embedded the pclntab inside `.data.rel.ro`, so this was correct. In Go 1.26, the linker restores `.gopclntab` as a standalone section for PIE builds. The pclntab magic is no longer in `.data.rel.ro`.

## Evidence (binary comparison)

We extracted `/usr/bin/skopeo` from two s390x RHEL 9 container images built in the same CI run:

| Binary | Go version | `.gopclntab` section | Magic in `.data.rel.ro` | check-payload result |
|--------|-----------|---------------------|------------------------|---------------------|
| skopeo 1.22.2-2.el9 (from `appstream`) | go1.26.1 (Red Hat 1.26.1-1.el9) | **Present** (magic `ff ff ff f1` at offset 0) | Not present | **FAIL** |
| skopeo 1.18.1-5.el9_6 (from `appstream-eus`) | go1.25.8 (Red Hat 1.25.8-1.el9_6) | Absent | **Present** (at offset 3278560) | **PASS** |

Both binaries are PIE, dynamically linked, stripped, s390x (big-endian). The only difference is the Go toolchain version used to build the RPM.

ELF section listing for the Go 1.26 binary:
```
[17] .gopclntab        PROGBITS  0000000000f60540 f60540 8f04e1 00   A  0   0 16
[23] .data.rel.ro      PROGBITS  000000000185ab00 1859b00 2eb418 00  WA  0   0 32
[28] .go.fipsinfo      PROGBITS  0000000001b7ce20 01b7be20  ...
[29] .go.module        PROGBITS  0000000001b7cea0 01b7bea0  ...
```

Note the new `.go.fipsinfo` and `.go.module` sections — consistent with Go 1.26's FIPS-related linker changes.

## Suggested Fix

Always check `.gopclntab` first, regardless of build mode. Fall back to `.data.rel.ro` only if `.gopclntab` is absent:

```go
section := exe.Section(".gopclntab")
if section == nil {
    section = exe.Section(".data.rel.ro")
    if section == nil {
        return nil, fmt.Errorf("could not read section .gopclntab from %s ", fileName)
    }
}
```

This is backward-compatible: Go <= 1.25 PIE binaries don't have `.gopclntab`, so the fallback to `.data.rel.ro` still applies.

## Related

- PR #164 — previous fix for s390x big-endian magic number handling (same code area)
- The RHEL 9 Go 1.26 toolchain (`golang-1.26.1-1.el9`) was recently rolled out to `appstream`; EUS repos still have Go 1.25.8
- `#forum-ocp-fips` discussion on Go 1.26 native FIPS support (`GOFIPS140=auto`): https://redhat-internal.slack.com/archives/C05U13J3LLS/p1772567553738529
- https://redhat.atlassian.net/browse/RHAIENG-4982

## Reproduction

```bash
# Pull an s390x image with skopeo built by Go 1.26
podman pull --platform linux/s390x registry.access.redhat.com/ubi9/ubi-minimal:latest
# Install skopeo 1.22.2 (from appstream, built with go1.26.1)
# Mount the image and run check-payload scan local --path <mount>
# Observe: "could not find magic number in .../usr/bin/skopeo"
```

cc @rphillips @smith-xyz

jiridanek:
## Summary

Go 1.26 changed the ELF layout for PIE builds: `.gopclntab` is now emitted as a
separate section instead of being embedded in `.data.rel.ro` (the Go <= 1.25
behavior). `ReadTable` unconditionally redirected to `.data.rel.ro` when it saw
`-buildmode=pie`, which breaks on Go 1.26 because `.data.rel.ro` no longer
contains the pclntab.

The impact depends on architecture:
- **s390x** (big-endian): fails deterministically — "could not find magic number"
- **amd64** (little-endian): appears to succeed, but the BE magic bytes
  (`ff ff ff f1`) match at a coincidental offset in `.data.rel.ro`. This produces
  an **empty symbol table** (0 functions), causing all subsequent Go FIPS
  validation (CGO, symbols, tags) to be **silently skipped** — no errors, but
  no actual validation either.

This fix always checks `.gopclntab` first, falling back to `.data.rel.ro` only
if absent. This is backward-compatible since Go <= 1.25 PIE binaries don't have
`.gopclntab`.

Fixes #329

## Evidence

Binary comparison of two s390x skopeo builds from RHEL 9 container images:

| Binary | Go version | `.gopclntab` section | Magic in `.data.rel.ro` | check-payload result |
|--------|-----------|---------------------|------------------------|---------------------|
| skopeo 1.22.2 (`appstream`) | go1.26.1 | Present (magic at offset 0) | Not present | FAIL |
| skopeo 1.18.1 (`appstream-eus`) | go1.25.8 | Absent | Present | PASS |

amd64 analysis of a Go 1.26 PIE binary:
- Fixed code reads `.gopclntab` → **66,602 functions** (correct)
- Buggy code reads `.data.rel.ro` at coincidental BE magic offset → **0 functions** (validation skipped)

## Tests

Added both unit and integration tests using a committed s390x PIE fixture binary
built with Go 1.26.3 (`GOFIPS140=certified`, cross-compiled with zig as linker):

- **`TestReadTable_NonPIE`**: existing arm64 non-PIE fixture — verifies the standard `.gopclntab` path
- **`TestReadTable_PIE_Go126_s390x`**: new s390x PIE fixture — regression test for #329
- **`TestRunLocalScan/PIE_Go126_s390x`**: full scan pipeline with the PIE fixture in a mock unpacked directory

With the buggy code reverted, the two s390x tests fail deterministically:
```
FAIL TestReadTable_PIE_Go126_s390x: could not find magic number in .../pie_go126_s390x_app
FAIL TestRunLocalScan/PIE_Go126_s390x: expected pass = true, got pass = false
```

s390x was chosen over amd64 for the fixture because on amd64 the old buggy code
produces an empty symbol table (via coincidental byte match) instead of an error,
so it would not catch the regression.

The fixture source (`test/resources/main.go`) and build script
(`test/resources/build_pie_fixture.sh`) are included for reproducibility.


smith-xyz (changes requested):
thank you for this fix! looks appropriate to me, but could we add a test to better document the flow here. I assume we haven't seen this, and we probably should work on some integration tests in ci to help with these kinds of issues.

smith-xyz (approved):
looks good - just merged some conflicts yesterday, can you rebase and I'll merge this in. Thank you for your contribution on this

jiridanek:
Good — the binary's structure is exactly what we need for the test. Now let me evaluate whether it's built "right" from the FIPS perspective based on dbenoit's guidance.

## Analysis of the fixture binary

The build metadata shows:

```
go1.26.3
-buildmode=pie
-tags=fips140v1.0
DefaultGODEBUG=fips140=on
CGO_ENABLED=1
GOFIPS140=v1.0.0-c2097c7c
GOARCH=s390x
```

**For the purpose of the test (check-payload section lookup fix): it's perfect.** The binary has `.gopclntab` as a separate section alongside `.data.rel.ro`, which is exactly the Go 1.26 PIE layout that triggers the bug.

**But from a FIPS compliance standpoint, there are two issues:**

1. **`GOFIPS140=certified` was used in the build script, but the binary shows `GOFIPS140=v1.0.0-c2097c7c`**. That's expected — `certified` is an alias that resolves to whatever version is actually certified. The binary records the resolved version. This is fine.

2. **Missing `GODEBUG=fips140=auto`**. The binary shows `DefaultGODEBUG=fips140=on`, not `fips140=auto`. According to dbenoit's rules:
   - `fips140=auto` means "read FIPS mode from the host" — this is what check-payload will enforce
   - `fips140=on` means "always on regardless of host"

   For our test fixture this **doesn't matter** — we're testing `ReadTable` section parsing, not the FIPS validation logic that checks for `fips140=auto`. The test just needs a binary with the right ELF layout.

3. **Cross-compiled with zig, not built on Red Hat's Go toolset**. The binary was compiled with upstream Go 1.26.3 (from `brew`/local install) cross-compiled via `zig cc`. It doesn't use Red Hat's `golang-fips/go` fork. Again, **irrelevant for the test** — `ReadTable` only cares about the ELF section layout, not who built the compiler.

## Verdict

**For its purpose as a test fixture: built correctly.** It triggers the exact regression we're testing for (`.gopclntab` separate from `.data.rel.ro` in a PIE binary on s390x big-endian).

**It is NOT a model of how to build a FIPS-compliant production binary.** But it was never intended to be — it's a minimal fixture (`func main() {}`) whose only job is to have the right ELF layout. The build script comments even acknowledge this:

```7:13:test/resources/build_pie_fixture.sh
# Uses zig as cross-linker (s390x PIE needs CGO/external linking).
# ...
# When ubi9/go-toolset:1.26 becomes available, consider rebuilding with:
#   podman run --platform linux/s390x -v $PWD:/out:Z ubi9/go-toolset:1.26 ...
```

If a reviewer asks "why not `fips140=auto`?": because we used `GOFIPS140=certified` which defaults to `fips140=on`, and setting `//go:debug fips140=auto` would require adding it to `main.go`. It doesn't affect the test at all, so there's no reason to complicate the fixture.

PR #331, opened 2026-05-09
chore: migrate .golangci.yml to v2 format

jiridanek:
## Summary

golangci-lint v2 requires a `version: "2"` field in the config file. Without it, the linter refuses to run:

```
Error: can't load config: unsupported version of the configuration: ""
See https://golangci-lint.run/docs/product/migration-guide for migration instructions
```

This was generated with `golangci-lint migrate`. Notable changes:
- `gofumpt` moved from `linters` to `formatters` (v2 distinguishes formatters from linters)
- `run.timeout` removed (disabled by default in v2)
- Default exclusion presets added by the migration tool

Note: one pre-existing `staticcheck` finding (`QF1008` in `internal/types/types_config.go:143`) is now surfaced. It was not previously visible because the v1 config suppressed it.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,12 +1,28 @@
 # For documentation, see https://golangci-lint.run/usage/configuration/
-
-run:
-  timeout: 5m
-
+version: "2"
 linters:
   enable:
-    - gofumpt
     - errorlint
+    - revive
     - unconvert
     - unparam
-    - revive
+  exclusions:
+    generated: lax
+    presets:
+      - comments
+      - common-false-positives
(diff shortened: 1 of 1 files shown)


smith-xyz (changes requested):
Is golangci-lint v2 actually available in CI? If not, this PR needs a paired change to the CI image/Makefile first.

As far as I can see its using v1.64.8

jiridanek:
I've been overly eager here, I think. I had v2 on my machine and I sort of went from there.

PR #333, opened 2026-05-11
fix: reject empty symbol table from ReadTable

jiridanek:
https://redhat.atlassian.net/browse/RHAIENG-4982

## Summary

Follow-up to #330. Add a guard in `_loadGoSymbols` to reject symbol tables
with zero functions.

A Go binary always has functions in its pclntab — even a trivial `func main() {}`
produces hundreds of runtime functions. An empty symbol table indicates the
pclntab was parsed from corrupt or unrelated data (e.g., a coincidental magic
byte match in the wrong ELF section).

## Problem

Without this guard, an empty table causes `isUsingCryptoModule` to return
`false` (no functions to iterate → no `crypto` found), setting `GoNoCrypto=true`.
This silently skips **all** subsequent FIPS validation:
- `validateGoSymbols` — skipped
- `validateGoCgo` — skipped
- `validateGoCGOInit` — skipped
- `validateGoStatic` — skipped
- `validateGoOpenssl` — skipped

The binary passes check-payload with no errors and no warnings, even though
no actual validation was performed.

## How we discovered this

During analysis for #329, we found that on amd64 Go 1.26 PIE binaries, the
old buggy section lookup read `.data.rel.ro` instead of `.gopclntab`. The BE
magic `ff ff ff f1` matched at a coincidental offset in `.data.rel.ro`,
producing an empty `gosym.Table` (0 funcs). The scan appeared to succeed
but performed zero Go FIPS validation.

## Change

```go
if len(symtable.Funcs) == 0 {
    return types.NewValidationError(fmt.Errorf(
        "go: symbol table for %v has no functions (pclntab may be corrupt)",
        filepath.Base(path)))
}
```

Made with [Cursor](https://cursor.com)

### DIFF
--- a/internal/validations/validations.go
+++ b/internal/validations/validations.go
@@ -97,6 +97,9 @@ func _loadGoSymbols(_ context.Context, path string, baton *Baton) *types.Validat
 	if err != nil {
 		return types.NewValidationError(fmt.Errorf("go: could not read table for %v: %w", filepath.Base(path), err))
 	}
+	if len(symtable.Funcs) == 0 {
+		return types.NewValidationError(fmt.Errorf("go: symbol table for %v has no functions (pclntab may be corrupt)", filepath.Base(path)))
+	}
 	baton.GoSymTable = symtable
 	if !isUsingCryptoModule(baton.GoSymTable) {
 		baton.GoNoCrypto = true


smith-xyz:
/lgtm

just doing some testing to ensure this doesn't break anything with various ocp versions

smith-xyz:
@jiridanek can you check this to see why the verify step is failing - it looks confused based on me trying to rebase your pr

ide-developer:
@smith-xyz The `ci/prow/verify` failure is not caused by this PR — it's a known regression in the CI infrastructure affecting multiple repos.

### TL;DR

[openshift/ci-tools#5161](https://github.com/openshift/ci-tools/pull/5161) by @Prucek (merged May 13) introduced sparse checkout for prow jobs. Test pods running from non-`src` containers (like our `verify` job which runs `from: golangci-lint`) now only get Dockerfiles checked out instead of the full source tree. A fix is in progress: [openshift/ci-tools#5179](https://github.com/openshift/ci-tools/pull/5179).

### Evidence

**This PR's failure:** `go generate ./internal/types` fails because `internal/types/` doesn't exist in the sparse checkout — only `Dockerfile` is checked out. The error is deterministic across 3 runs with different commit hashes. Our change only adds 3 lines to `internal/validations/validations.go`.

**Other repos also broken:** The same lint failures are being `/override`'d on `openshift/ci-tools` itself:
- [ci-tools#5183](https://github.com/openshift/ci-tools/pull/5183#issuecomment-4444152213) — lint overridden by @deepsm007
- [ci-tools#5182](https://github.com/openshift/ci-tools/pull/5182#issuecomment-4443167429) — lint overridden by @deepsm007
- [ci-tools#5181](https://github.com/openshift/ci-tools/pull/5181#issuecomment-4445539962) — lint overridden by @jmguzik

**Why it broke:** The [regenerated prow job YAML](https://github.com/openshift/release/commit/a98c22f57a2a) for check-payload sets `sparse_checkout_files: ["Dockerfile"]` on all jobs including `verify`. The `src` image build clears this for a full clone, but non-`src` containers like `golangci-lint` get the sparse checkout directly from prow's clonerefs.

```yaml
# check-payload's generated job YAML (all jobs, including verify):
decoration_config:
  sparse_checkout_files:
  - Dockerfile
```

**Why PR #334 passes:** It ran on May 12 ([log](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_check-payload/334/pull-ci-openshift-check-payload-main-verify/2054285965886754816)) with `ci-operator v20260512`, before the sparse checkout was deployed. Once #334 is retested, it will likely fail too.

### Resolution

Once [ci-tools#5179](https://github.com/openshift/ci-tools/pull/5179) merges (clears `SparseCheckoutFiles` for test pods), our `verify` should pass. In the meantime, `/override verify` is the same workaround the ci-tools team is using on their own PRs.

Alternatively, adding `clone: true` to the verify job in [openshift/release](https://github.com/openshift/release/blob/main/ci-operator/config/openshift/check-payload/openshift-check-payload-main.yaml) would also fix it (this is what repos like installer, insights-operator, kube-compare already have on their golangci-lint tests).

smith-xyz:
thank you @jiridanek for finding this one !

PR #338, opened 2026-05-20
feat: containerized fixture builder and cross-arch test coverage

bartoszmajsak:
## Summary

Follow-up to #337. Replaces ad-hoc per-fixture build scripts with a single parameterized builder backed by Red Hat's go-toolset container images and zig for cross-compilation. All test fixtures are now built from one `main.go`, producing consistent FIPS settings across Go versions.

Addresses #337 review feedback:
- Fixture source/binary divergence - unified source, reproducible builds
- Missing build scripts - single parameterized `build_fixture.sh`
- No integration tests for new fixtures - added for all fixture variants
- Swallowed `tryParseTable` errors - `slog.Debug` for rejected candidates
- Test name inaccuracies - s390x fixture is CGO not PIE

Fills two coverage gaps:
- **ppc64le** - quantum=4 + little-endian, previously untested
- **`.data.rel.ro` fallback** - Go 1.24 external PIE exercises the magic scanning loop that no other fixture reaches

Adds fixture invariant assertions so rebuilt binaries can't silently change which code path they exercise.

## Test plan

- [ ] Rebuild from scratch: remove builder images, run `build_fixture.sh` per README, commit, `make test verify`
- [ ] `make test` passes (135 tests across 9 packages)
- [ ] Fixture invariant assertions catch wrong section/arch if a binary is rebuilt incorrectly


bartoszmajsak:
@smith-xyz PTAL when you find some time.

smith-xyz:
@bartoszmajsak LGTM. Before merge, can we add rejectSections for go124_external_pie_amd64_app so we don’t regress the .data.rel.ro scan path silently; consider TestReadTable for go-native-fips-app and t.Fatal in assertFixtureInvariants.

bartoszmajsak:
> @bartoszmajsak LGTM. Before merge, can we add rejectSections for go124_external_pie_amd64_app so we don’t regress the .data.rel.ro scan path silently; consider TestReadTable for go-native-fips-app and t.Fatal in assertFixtureInvariants.

@smith-xyz Good feedback, thanks! Adressed in 309c7a763ae5c4e16cc69f34906f8bccff2043db

>  consider TestReadTable for go-native-fips-app

IIUC `ReadTable` doesn't care about the crypto stack - it parses pclntab, which is the same format regardless of whether the binary uses boringcrypto or crypto/fips140. It's a bit of a redundant test, but cheap, so I added it.

smith-xyz:
@bartoszmajsak can you rebase and resolve the conflict so we can merge this in?

PR #339, opened 2026-05-20
The MCE hypershift container needs changes to the exceptions in place

yiraeChristineKim:
The FIPS scans are failing in our konflux builds due to problems with the hypershift "extra" binaries.  The current exceptions for these binaries is not correctly preventing the scanner from erroring on these binaries. Tested with the latest image from our MCE 2.18.3 release
from https://github.com/openshift/check-payload/pull/282
Refs:
 - https://issues.redhat.com/browse/ACM-24281
 - https://redhat.atlassian.net/browse/ACM-34234

### DIFF
--- a/dist/releases/4.15/config.toml
+++ b/dist/releases/4.15/config.toml
@@ -178,7 +178,23 @@ files = [
 
 [[payload.multicluster-engine-hypershift-operator.ignore]]
 error = "ErrGoNotCgoEnabled"
-files = ["/usr/bin/hcp-no-cgo", "/usr/bin/hypershift-no-cgo"]
+files = ["/usr/bin/hcp", "/usr/bin/hcp-no-cgo", "/usr/bin/hypershift-no-cgo"]
+
+[[payload.multicluster-engine-hypershift-operator-container.ignore]]
+error = "ErrGoMissingSymbols"
+files = [
+  "/usr/bin/hcp",
+  "/usr/bin/hcp-no-cgo",
+  "/usr/bin/hypershift-no-cgo",
+]
+
+[[payload.multicluster-engine-hypershift-operator-container.ignore]]
+error = "ErrNotDynLinked"
+files = [
+  "/usr/bin/hcp",
+  "/usr/bin/hcp-no-cgo",
+  "/usr/bin/hypershift-no-cgo",
+]
 
 [[payload.virt-cdi-importer-rhel9-container.ignore]]
 error = "ErrGoNotCgoEnabled"
--- a/dist/releases/4.16/config.toml
+++ b/dist/releases/4.16/config.toml
@@ -462,7 +462,23 @@ files = ["/tools/opm-rhel8"]
 
 [[payload.multicluster-engine-hypershift-operator.ignore]]
 error = "ErrGoNotCgoEnabled"
-files = ["/usr/bin/hcp-no-cgo", "/usr/bin/hypershift-no-cgo"]
+files = ["/usr/bin/hcp", "/usr/bin/hcp-no-cgo", "/usr/bin/hypershift-no-cgo"]
+
+[[payload.multicluster-engine-hypershift-operator-container.ignore]]
+error = "ErrGoMissingSymbols"
+files = [
+  "/usr/bin/hcp",
+  "/usr/bin/hcp-no-cgo",
+  "/usr/bin/hypershift-no-cgo",
+]
+
+[[payload.multicluster-engine-hypershift-operator-container.ignore]]
+error = "ErrNotDynLinked"
+files = [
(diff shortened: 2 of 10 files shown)


smith-xyz (changes requested):
A few questions:

1. Why can't the -no-cgo and hypershift-no-cgo binaries be removed from the production image? They're documented as QE-only.
2. Why is the primary /usr/bin/hcp binary now non-CGO/non-FIPS? Is there a plan to fix this upstream rather than permanently excepting it?
3. If exceptions are truly necessary (e.g., short-term while upstream fixes land), they need justification comments, bug links, and ideally a time-boxed plan to remove them.

Wondering if this is due to a split here between konflux and ART? Modifying here means that when ART runs this against the release payload, the exceptions trickle up.

If this is purely at some level with konflux then is there a way to have some override at the konflux level?

Does ART actually strip /usr/bin/hcp, /usr/bin/hcp-no-cgo, and /usr/bin/hypershift-no-cgo from the final image?

gparvin:
> A few questions:
> 
>     1. Why can't the -no-cgo and hypershift-no-cgo binaries be removed from the production image? They're documented as QE-only.

Right, QE is requiring these files be included.
 
>     2. Why is the primary /usr/bin/hcp binary now non-CGO/non-FIPS? Is there a plan to fix this upstream rather than permanently excepting it?
> 

When we move to ART, we can just not include the binaries at all.  It's not clear exactly when this will be, but I expect it will not be more than a few months for all streams to move over.

>     3. If exceptions are truly necessary (e.g., short-term while upstream fixes land), they need justification comments, bug links, and ideally a time-boxed plan to remove them.
> 

The QE team doesn't have a plan to move away from using these binaries, but we should be able to include them in ACM builds that don't scan for FIPS and not include them in ART builds -- when we move to ART.  For now, our ACM dev pipeline and production pipelines use the same dockerfile so the binaries are included.

> 
> Wondering if this is due to a split here between konflux and ART? Modifying here means that when ART runs this against the release payload, the exceptions trickle up.
> 
> If this is purely at some level with konflux then is there a way to have some override at the konflux level?
> 
> Does ART actually strip /usr/bin/hcp, /usr/bin/hcp-no-cgo, and /usr/bin/hypershift-no-cgo from the final image?

I'm not sure, I was thinking our ART dockerfile could not include the binaries at all.  I'll suggest this to the hypershift team to see if it would work.

rokej:
@gparvin Thank you for answering those questions. Let me know if further clarifications are needed.

yiraeChristineKim:
@smith-xyz  Can we merge this? It’s been pending for a long time.

smith-xyz:
@yiraeChristineKim @gparvin Thanks, this helps, but it reinforces why I’m hesitant to merge as-is.

Release configs ≠ Konflux workaround. dist/releases/*/config.toml is used when scanning the OCP release payload. If ART omits these binaries, we shouldn’t need exceptions here. If ACM Konflux needs a pass, that belongs in the Konflux pipeline / scan invocation, not in release-wide exceptions.

Same Dockerfile for ACM dev and prod means QE-only binaries are shipping to ACM customers today. Widening exceptions doesn’t address that; splitting the Dockerfile (or a build arg to omit -no-cgo / non-FIPS hcp from prod) does.

/usr/bin/hcp is the production binary. Excepting it for ErrGoNotCgoEnabled / missing symbols / static link is accepting non-FIPS crypto for the main tool. “We’ll drop it in ART later” isn’t a substitute for either fixing the build or not shipping it now.

Before merge: confirm what com.redhat.component is on Konflux vs ART images; confirm whether these paths exist in the composed OCP payload; add justification comments + [ACM-24281](https://redhat.atlassian.net/browse/ACM-24281) in the TOML; scope exceptions to releases that actually ship this image.

Happy to re-review if the approach is “fix Dockerfile / omit binaries from prod” or a Konflux-specific config path — less happy to merge broad release exceptions for binaries you plan to stop shipping.

smith-xyz:
just thinking of a few cleaner paths we can take in case this can help - I'm not sure what is possible relative to konflux here but just to see if we can move this along:

1. Can we do something like this for ACM Konflux pipeline: check-payload … --config ./hack/mce-fips-exceptions.toml (and only -V if you still need base policy). Keeps ACM policy in ACM repos. If you're able to override the task somehow maybe that is a path forward.
2. Split Dockerfile / build targets: prod image without -no-cgo / non-FIPS hcp; QE image or separate artifact - this is more or less what we've mentioned will happen with ART so splitting it now would make this simpler.

yiraeChristineKim:
@smith-xyz  Currently we are trying to release hypershift hcp cli to Developer portal, then If we omit binaries from prod, then we cannot publish to developer portal .... right?

PR #340, opened 2026-05-27
Extend virt-launcher ErrNotDynLinked exception to new virt-launcher for nvidia

Axuba:
The new virt-launcher container that we produce is built like the original virt-launcher that we already have an exception approved for ErrNotDynLinked. The only difference is that it's based on rhel 10.2 to use a libvirt and qemu versions with nvidia patches.


smith-xyz (approved):
/lgtm
/approve

Issue #342, opened 2026-06-17
Native Go FIPS binaries (GOFIPS140 + no_openssl) still flagged after #335


hbelmiro:
## Summary

Binaries built with Go native FIPS (`GOFIPS140=v1.0.0 -tags no_openssl`) are still flagged by check-payload at latest commit (4dda1bfb, post-#335 merge).

## Reproduction

Build and push this image:

```dockerfile
FROM registry.access.redhat.com/ubi9/go-toolset:1.26.3 AS builder
USER root
WORKDIR /workspace
RUN echo 'package main; import "fmt"; func main() { fmt.Println("hello") }' > main.go
RUN go mod init test
RUN GOTOOLCHAIN=local CGO_ENABLED=0 GOFIPS140=v1.0.0 \
    go build -tags netgo,no_openssl -o /bin/test-binary .

FROM registry.access.redhat.com/ubi9/ubi-minimal:9.5
ENV GODEBUG=fips140=on
COPY --from=builder /bin/test-binary /usr/bin/test-binary
ENTRYPOINT ["/usr/bin/test-binary"]
```

Then scan:
```bash
check-payload scan operator --spec <image>
```

## Results

**CGO_ENABLED=0** (native FIPS target):
```
go binary is not CGO_ENABLED
```

**CGO_ENABLED=1 -tags no_openssl** (attempted workaround):
```
go binary does not contain required symbol(s)
```

Both fail because check-payload looks for OpenSSL symbols/linkage, which don't exist in native Go FIPS binaries.

## Context

- Go native FIPS module is [CMVP certified (#5247)](https://csrc.nist.gov/projects/cryptographic-module-validation-program/certificate/5247)
- Red Hat go-toolset supports native FIPS (GOFIPS140) [since 1.26](https://github.com/golang-fips/go?tab=readme-ov-file#native-fips-module-support)
- golang-fips/go fork states Go 1.26 is the [final release with OpenSSL backend](https://github.com/golang-fips/go?tab=readme-ov-file#migration-to-upstream-fips-certified-cryptography)
- `-tags no_openssl` and GOFIPS140 are mutually exclusive with the OpenSSL backend — we cannot have both

## Related

- #332 (closed by #335, but the fix doesn't cover this case)

hbelmiro:
Fixed by https://github.com/openshift/check-payload/pull/344

PR #344, opened 2026-06-18
fix(fips): accept fips140=on and fips140=only as valid native FIPS modes

smith-xyz:
The golang fips check is too narrow, so this will widen `hasGodebugFIPS140Auto` as `hasGodebugFIPS140Enabled` to accept `fips140=on` and `fips140=only` in addition to `fips140=auto`. The RH Go 1.26 toolchain defaults to `fips140=on` when `GOFIPS140` is set, which was incorrectly rejected.


smith-xyz:
Fixes #342

PR #352, opened 2026-07-30
FROM scratch data-only image containing skill definition files so the…

pruan-rht:
…re's no OS layer thus FIPS OS validatioin is not applicable

rh-pre-commit.version: 2.3.2
rh-pre-commit.check-secrets: ENABLED

https://redhat-internal.slack.com/archives/CB95J6R4N/p1785445793383609


sdodson:
/label approved

Approving the notion that we can have this image in the payload because it contains no binary content. Need to understand why jobs are still failing so asking @smith-xyz to review

PR #356, opened 2026-08-05
fix(scan): classify missing release file and add --tag to local scans

smith-xyz:
## Summary

- Classify a missing `/etc/redhat-release` detected via `Lstat` as
  `ErrDistributionFileMissing` instead of leaking a raw lstat error that no
  config exception can match (previously only the `ReadFile` path was
  classified).
- Add a `--tag` flag to `scan local` so tag-scoped config exceptions (e.g.
  `[[tag.agentic-skills.ignore]]`) can apply. Local scans previously used an
  empty mock tag name, making every tag ignore unreachable. The tag carries
  identity only — the scan root remains `--path` and is never resolved to a
  `<path>/<tag>` subdirectory (pinned by test).
- Fixes the `payload-scan-5-0` smoke test failure on the `agentic-skills`
  FROM-scratch data-only image: no OS layer, zero binaries, and the existing
  `ErrOSNotCertified` tag ignore could not fire under `scan local`.

Without `--tag`, behavior is unchanged: the mock tag name stays empty, no
ignores match, and the scan root normalization (`filepath.Clean`) is identical
to the previous `filepath.Join(path, "")`.

Companion openshift/release change passes the tag name per image in the
`payload-scan-*` steps; this PR must merge first since old binaries reject the
new flag.

### DIFF
--- a/internal/scan/scan.go
+++ b/internal/scan/scan.go
@@ -130,7 +130,7 @@ func RunLocalScan(ctx context.Context, cfg *types.Config, localBundlePath string
 	var runs []*types.ScanResults
 
 	// Simulate payload based on local directory structure
-	localPayload := simulateLocalPayload(localBundlePath)
+	localPayload := simulateLocalPayload(localBundlePath, cfg.LocalTag)
 
 	// Rest of the function follows the structure of RunPayloadScan
 	parallelism := cfg.Parallelism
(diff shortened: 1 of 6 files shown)


smith-xyz:
/override ci/prow/payload-scan-5-0

this is to fix that issue, will need a change in openshift/release

PR #284, opened 2025-10-30
Add exception for ripgrep in Dev Spaces Code

SDawley:
Hey folks, I'd like to get an exception added for Dev Spaces if possible.

Justification: ripgrep (specifically [microsoft/ripgrep-prebuilt](https://github.com/microsoft/ripgrep-prebuilt)) is a requirement for VS Code, which is built and used in Dev Spaces Code. These are not dynamically linked, but I have not seen anything to indicate that ripgrep does anything involving cryptography outside of allowing the user to specify something else as a preprocessor.

Please let me know if you have any questions, or I've formatted something incorrectly, or if I can just put this in the top level config.toml.

Thanks

PR #357, opened 2026-08-14
Add exception for ripgrep-universal in Dev Spaces Code

RomanNikitenko:
## Context
-  Dev Spaces has an exception for `@vscode/ripgrep`, see https://github.com/openshift/check-payload/pull/284
- Starting with Dev Spaces `3.30`, the upstream VS Code replaced `@vscode/ripgrep` with `@vscode/ripgrep-universal` as a production dependency.

`@vscode/ripgrep-universal` is a new package from the same repository (microsoft/vscode-ripgrep) that bundles ripgrep (BurntSushi/ripgrep) binaries for all platforms in a single tarball, eliminating the `postinstall` download step.
Non-target platform binaries are stripped at VS Code packaging time.
See: https://github.com/microsoft/vscode/commit/c4471e24faf21ae50af048d09d68cb6e3092e4ad

The binary path changed from `node_modules/@vscode/ripgrep/bin/rg` to `node_modules/@vscode/ripgrep-universal/bin/<platform>-<arch>/rg`.

## Justification

ripgrep (BurntSushi/ripgrep) is a text search tool. It does not perform cryptographic operations.

Verified by inspecting the binaries in the container image `quay.io/redhat-user-workloads/devspaces-tenant/devspaces/code-rhel9:3.30` across all four architectures (amd64, arm64, s390x, ppc64le):
- `strings <binary> | grep -iE` with word-boundary matching found only false positives:
  "rsa" as a Unicode script code (ISO 15924) embedded in Rust's regex engine,  "TLS" referring to Thread-Local Storage in Rust runtime
- Zero matches for actual cryptographic APIs or libraries: openssl, libssl, libcrypto, fips, cipher, encrypt, decrypt, EVP_*, PKCS, x509, certificate, hmac
  
## Changes

- Adds `@vscode/ripgrep-universal` paths for all supported architectures (linux-x64, linux-arm64, linux-ppc64, linux-s390x) and both UBI variants (ubi8, ubi9) to the `devspaces-code-rhel9-container` exception in releases:
4.16, 4.18, 4.19, 4.20, 4.21, 4.22, 4.23, 5.0, 5.1.
- The old `@vscode/ripgrep/bin/rg` paths are preserved for backward compatibility.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -252,6 +252,36 @@ error = "ErrNotDynLinked"
 files = ["/usr/local/bin/mta-cli"]
 
 # -----------------------------------------------------------------------------
+# Dev Spaces Code (devspaces-code-rhel9-container)
+#
+# ripgrep (BurntSushi/ripgrep) is a text search tool used by VS Code for
+# workspace search. It does not perform any cryptographic operations.
+# Binary inspection confirmed zero references to OpenSSL, libcrypto, FIPS APIs,
+# or any cipher/TLS primitives — only false positives from Unicode script codes
+# and Rust Thread-Local Storage symbols.
+#
+# @vscode/ripgrep — original single-platform package (legacy, kept for
+# backward compatibility with older builds).
+# @vscode/ripgrep-universal — new all-platform bundle introduced upstream;
+# non-target platform binaries are stripped at VS Code packaging time.
+# See: https://github.com/microsoft/vscode/commit/c4471e2
+# -----------------------------------------------------------------------------
+[[payload.devspaces-code-rhel9-container.ignore]]
+error = "ErrNotDynLinked"
+files = [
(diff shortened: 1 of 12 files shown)


smith-xyz:
@RomanNikitenko , if this is an expected tool irregardless of ocp versions, it may be better to place it in the global config.toml ?

Also, if you can just note the justification above it on why its there, why it doesn't violate fips (no crypto usage), and why it is necessary for customers to receive this in the bundle that would be most appreciated.

RomanNikitenko:
@smith-xyz 
Good point! 
I've moved the exception to the global `config.toml` with a detailed justification comment explaining:

- ripgrep is a text search tool with no cryptographic operations (confirmed via binary inspection)
- the difference between `@vscode/ripgrep` (legacy) and `@vscode/ripgrep-universal` (new multi-platform bundle)
- upstream commit reference

PR #360, opened 2026-09-15
feat(validations): add Rust binary FIPS support

### LINKED ISSUE #359: Add Rust binary support
Rust is becoming a significant part of Red Hat's shipped products and tooling, and the FIPS scanners need to follow. check-payload has a dedicated Go pipeline and a generic ELF pipeline but no Rust arm, so Rust binaries go effectively unchecked.

**Problem**

check-payload has no Rust classifier, so Rust binaries fall through to the generic exe pipeline, which fails only static binaries and never requires linkage to the system OpenSSL. A Rust binary can do crypto through a vendored, static, or pure-Rust backend such as ring, aws-lc-rs, vendored openssl-sys, or RustCrypto like sha2. Any of those passes the scan without reaching the validated module. Today Rust binaries surface only as one-off exceptions, for example ripgrep in #357, never as a validated crypto path.

**Why it matters**

Rust is arriving in OpenShift payloads and operator images. FIPS requires every crypto operation to run in the system OpenSSL FIPS provider, dynamically linked with nothing vendored. check-payload cannot yet tell a compliant Rust binary from a non-compliant one.

**Proposed direction**

No single check is sufficient, so a Rust classifier feeds a pipeline that layers three:

- dynamic linkage to the system libcrypto, proving the binary reaches the validated module,
- a symbol-table scan for self-contained backends like ring or vendored OpenSSL, needing no SBOM,
- the cargo auditable SBOM, for pure-Rust crypto the symbols cannot see.

**Known limit**

Each layer covers a gap the others leave. ELF inspection misses pure-Rust crypto like sha2. It exposes no stable C-ABI symbol to fingerprint, and release builds strip its mangled Rust symbols. That class rests on the cargo auditable SBOM embedded at build time. A Hermeto or Konflux SBOM does not substitute, since it comes from Cargo.lock and over-reports crates the build never compiled (https://github.com/orgs/hermetoproject/discussions/1769).

Happy to work up the detail if useful.

hexfusion:
## What

Adds a Rust arm to the scanner so Rust ELF binaries are checked for FIPS compliance instead of falling through the generic executable pipeline.

Closes #359.

## How

Rust binaries are detected by their cargo-auditable section or the `rustc` producer string, then checked with two independent signals corroborated so neither is trusted alone:

- A symbol scan for a bundled crypto backend (ring, a vendored or static OpenSSL, BoringSSL, or aws-lc), matched by defined symbols rather than by crate name. Undefined imports that resolve to the system libcrypto are the compliant case and are not flagged.
- The cargo-auditable manifest, matched against a configurable `rust_denied_crypto` denylist, which also catches the pure-Rust primitives that leave no symbol.

Either signal fails the binary. A binary with no manifest and no bundled-backend symbol is reported as a warning, because a pure-Rust primitive cannot be ruled out from the binary alone. Building Rust binaries with `cargo auditable` gives the manifest the second signal relies on.

The denylist is config-driven (`rust_denied_crypto` in `config.toml`) with a built-in fallback, and is a best-effort, non-exhaustive supplement to the structural symbol scan. Details are in the README.

## Testing

Unit tests cover the symbol classification, the defined-vs-undefined skip, the manifest parse and deny-matching, the corroboration decision, and the config setter. `go build`, `go vet`, `gofmt`, and `make test` pass on go 1.26.

### FIPS VM validation

Validated end to end on a FIPS-enabled RHEL VM, since the payload-scan CI is currently blocked by an unrelated GSM/CSI credential-migration issue and does not exercise this code.

Environment: RHEL 9.7, FIPS mode on (`/proc/sys/crypto/fips_enabled` = 1), system OpenSSL 3.5.8, cargo 1.98.1, Go 1.26.0 (official tarball; go.mod requires 1.26.0).

Build and unit tests:

```
git clone https://github.com/hexfusion/check-payload -b feat/rust-support
cd check-payload
make all      # CGO_ENABLED=0
make test     # PASS, including internal/rust and internal/validations
```

Fixtures:
- ring-bad: `ring = "0.17"` (calls ring::digest). Built two ways: `cargo auditable build --release` (carries the `.dep-v0` SBOM plus defined `ring_core_*` symbols) and plain `cargo build --release` (no `.dep-v0`, `ring_core_*` symbols present, recognized as Rust via the rustc string in `.comment`).
- openssl-good: `openssl = "0.10"` without the vendored feature (calls openssl::hash). `ldd` confirms it dynamically links the system `libssl.so.3` / `libcrypto.so.3`; `nm -D` shows the `OPENSSL_`/`EVP_` symbols are UNDEF imports with no defined bundled-crypto symbols. Built with cargo-auditable.

Scans (`check-payload scan local --path <dir> -v=1`):
- ring-bad (cargo-auditable): FLAGGED. "rust binary bundles unvalidated crypto instead of the system OpenSSL provider (symbols: ring; manifest: ring)". Both signals fired.
- ring-bad (plain, no `.dep-v0`): FLAGGED. "(symbols: ring)". Symbol path only, as expected with no manifest.
- openssl-good: PASSED the Rust check (binary scanned status success; the bundled-crypto validation did not flag it).

Note: the openssl-good run exits non-zero on a separate, pre-existing module-level check ("no FIPS certified artifact or library found for module openssl"), because a bare binary was scanned outside an image that ships the openssl RPM. That check is orthogonal to this PR; if anything it confirms the binary took the system-OpenSSL path, since a bundled-crypto binary never triggers the openssl module requirement.

Result: the Rust FIPS detection flags bundled crypto (ring, via both the cargo-auditable SBOM and defined symbols) and passes system-linked OpenSSL, with no false pass or false fail.


README.md
```diff
@@ -174,6 +174,17 @@ JDK validations run through a pipeline:
 1. validateSystemProperties - ensures pertinent [FIPS property values](https://access.redhat.com/documentation/en-us/openjdk/8/html/configuring_openjdk_8_on_rhel_with_fips/config-fips-in-openjdk) are not being set at runtime
 1. validateAlgorithms - ensures unacceptable algorithms and protocols are disabled at runtime
 
+#### Rust Executables
+
+Rust executables are detected by their cargo-auditable section or the `rustc` producer string, then checked for crypto that runs outside the system OpenSSL FIPS provider. Two independent signals are corroborated so neither is trusted alone:
```
smith-xyz: 👍 good call

config.toml
```diff
@@ -17,6 +17,19 @@ certified_distributions = [
   "Red Hat Enterprise Linux release 10.2 (Coughlan)",
 ]
 
+# Rust crates that provide crypto outside the system OpenSSL FIPS provider,
+# chiefly the pure-Rust primitives the symbol scan cannot see. Best-effort and
+# non-exhaustive. Extend here rather than in code.
+rust_denied_crypto = [
+  # bundled or alternative TLS/crypto stacks
+  "ring", "rustls", "aws-lc-rs", "aws-lc-sys", "aws-lc-fips-sys",
```
smith-xyz: Should this be unconditionally denied? aws-lc-fips-sys can represent a validated static AWS-LC module, depending on its exact version, build, and operating environment. Denying it rejects a potentially valid module; removing it from this list would accept it without validating those details. I think this should map to a stable module ID and flow through fips_certified_modules, with the resolved crate/module version retained as evidence.

smith-xyz: I do not think rustls itself is a reliable deny signal. It separates TLS from its CryptoProvider and can use ring, AWS-LC FIPS, or an OpenSSL-backed provider. A Rustls build using system OpenSSL would still fail here based only on the protocol crate name. Could we classify the selected provider instead?

internal/validations/validations.go
```diff
@@ -80,9 +81,36 @@ type Baton struct {
 	GoVersion   *semver.Version
 	GoBuildInfo *buildinfo.BuildInfo
 	GoSymTable  *gosym.Table
+	RustAudit   *rust.Sbom
 	ModulesUsed []string
 }
 
+// defaultDeniedRustCrypto is the fallback when config omits rust_denied_crypto.
+var defaultDeniedRustCrypto = map[string]struct{}{
```
smith-xyz: ... I think I don't prefer a fallback here? If we do that for Go its fine then, just unnecessary bloat when the config is meant to define the thing - error if it isn't there.

internal/validations/validations.go
```diff
@@ -100,6 +128,11 @@ var validationFns = map[string][]ValidationFn{
 		validateNotStatic,
 		validateExeOpenssl,
 	},
+	"rust": {
+		validateNotStatic,         // a static Rust binary carries its own crypto
```
smith-xyz: Is rejecting every static Rust binary an intentional OpenShift support policy? Static linking is not inherently non-FIPS: AWS-LC has FIPS-validated static modules intended for embedding. If the product policy is “only RHEL system crypto is supported,” I think we should state that explicitly. Otherwise this should validate the embedded module rather than reject static linkage itself.

internal/validations/validations.go
```diff
@@ -456,6 +489,59 @@ func isGoExecutable(path string, baton *Baton) (bool, error) {
...
+	if err != nil {
+		return false, nil
+	}
+	defer exe.Close()
+
+	if !rust.IsRustExecutable(exe) {
+		return false, nil
+	}
+	// A Rust binary without cargo-auditable is still Rust. The validator warns.
+	if sbom, err := rust.ReadAuditable(exe); err == nil {
```
smith-xyz: This drops a malformed .dep-v0 error and later reports only “no auditable manifest” as a warning. That means corrupt or deliberately invalid audit evidence can turn into a successful scan? Can we distinguish an absent section from a present-but-invalid section and fail closed on parse errors?

Keep me honest if that is wrong.

internal/rust/rust.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+// ParseAuditable decodes the zlib-JSON of a cargo-auditable section.
+func ParseAuditable(compressed []byte) (*Sbom, error) {
+	zr, err := zlib.NewReader(bytes.NewReader(compressed))
+	if err != nil {
+		return nil, err
+	}
+	defer zr.Close()
+	decompressed, err := io.ReadAll(zr)
```
smith-xyz: We consider the payload as untrusted scanner input, could we cap decompressed output? Upstream cargo-auditable recommends an 8 MiB limit to guard against zip bombs.

internal/rust/rust.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+// LinkedCrypto returns denied crypto crates linked in, ignoring build and dev deps.
+func (s *Sbom) LinkedCrypto(denied map[string]struct{}) []string {
+	if s == nil {
+		return nil
+	}
+	var found []string
+	for _, p := range s.Packages {
+		if p.Kind != "" && p.Kind != "runtime" {
```
smith-xyz: openssl-src is normally a build dependency because it compiles OpenSSL objects that are then linked into the runtime binary. 

Does ignoring all build dependencies defeat the explicit openssl-src denylist entry. Could native source crates be checked regardless of Cargo dependency kind?

smith-xyz (changes requested):
I think I have more questions than anything - this is a nice implementation to help detect crypto usage from rust binaries. 

My main concern is the distinction between detecting suspicious crypto and proving use of an approved module. This PR does the former reasonably well, but absence from a best-effort crate denylist is not positive FIPS evidence. I would be more comfortable if this PR reported provider/module candidates and reused the generic certified-module validation path.

config.toml
hexfusion: it's a validated module at specific versions. i could resolve it through fips_certified_modules with version bounds instead of denying by name. good idea.

hexfusion: good call, I'll drop it and classify on the provider, which is looks like its visible in SBOM/symbols.

internal/validations/validations.go
hexfusion: yeah same check the exe/Go arms use, but you're right about validated static. 

> If the product policy is “only RHEL system crypto is supported,” I think we should state that explicitly. 

we are looking to the crypto/security team for guidance here, my impression was yes for now.

internal/validations/validations.go
hexfusion: great catch! this looks like would fail open, i will close that up and fail closed if we hit an unparseable section and warn if absent.

hexfusion:
> I think I have more questions than anything - this is a nice implementation to help detect crypto usage from rust binaries.
> 
> My main concern is the distinction between detecting suspicious crypto and proving use of an approved module. This PR does the former reasonably well, but absence from a best-effort crate denylist is not positive FIPS evidence. I would be more comfortable if this PR reported provider/module candidates and reused the generic certified-module validation path.

great feedback and I agree. i reworked it so the denylist is only detector input, not the actual pass/fail authority. now detected providers are reported as module candidates and validated through the same fips_certified_modules path as the openssL and golang arms, so a binary passes only on positive attestation. one caveat i left as a follow-up, this attests a module by name, not by cmvp-validated version, since defining certified-version semantics for a statically bundled module felt like your teams call to make. but if you have guidance i can apply now.

internal/validations/validations.go
```diff
@@ -456,6 +493,64 @@ func isGoExecutable(path string, baton *Baton) (bool, error) {
...
+			unattested = append(unattested, m)
+		}
+	}
+	if len(unattested) > 0 {
+		return attested, types.NewValidationError(fmt.Errorf("%w (%s)", types.ErrRustBundledCrypto, strings.Join(unattested, ", ")))
+	}
+	if len(candidates) == 0 && !hasManifest {
+		return nil, types.NewValidationError(types.ErrRustNoAuditable).SetWarning()
+	}
+	return attested, nil
```
smith-xyz: Could this return indeterminate rather than success when a manifest exists but produces no candidates?

Concrete flow: `.dep-v0` lists `private-crypto` (or any unlisted crate) → `LinkedCrypto` returns `[]` → `candidates` is empty → the warning at line 548 is skipped because `hasManifest == true` → this returns `nil`, so the binary passes. The existing `clean with manifest passes` test codifies this.

Because `rust_denied_crypto` is explicitly non-exhaustive, no denylist match is not positive provider evidence. I think this should warn/fail as indeterminate unless another signal identifies an approved provider.

internal/scan/node_scan.go
```diff
@@ -13,6 +13,8 @@ import (
 )
 
 func RunNodeScan(ctx context.Context, cfg *types.Config, root string) []*types.ScanResults {
+	validations.SetRustDeniedCrypto(cfg.RustDeniedCrypto)
+	validations.SetRustCertifiedModules(cfg.GetFIPSCertifiedModules())
```
smith-xyz: This installs certified module **names**, but the RPM branch below calls `rpmRootScan` and never runs `validateModuleArtifactsPhase`.

Concrete flow: Rust candidate matches an image-source module name → inline classification passes and records `ModulesUsed` → `rpmRootScan` returns → a missing or out-of-range configured artifact is never checked. `walkDirScan` does perform that check.

This appears to be a broader existing node/RPM pipeline limitation, not only a Rust issue. Could we either run the generic module phase over RPM results or document that RPM node mode only name-gates modules?

internal/rust/rust.go
```diff
@@ -0,0 +1,195 @@
...
+	"bundled-openssl": {"openssl-src"},
+	"ring":            {"ring"},
+}
+
+// CryptoModuleCandidates names the bundled crypto providers a binary carries, as
+// module names the certified-module path can attest. Manifest crate names win over
+// the coarser symbol backends.
+func CryptoModuleCandidates(sbom *Sbom, denied map[string]struct{}, symbolBackends []string) []string {
+	crates := sbom.LinkedCrypto(denied)
+	out := slices.Clone(crates)
```
smith-xyz: Could this collapse provider-family evidence instead of treating every matching crate as a separate certified module?

Concrete FIPS graph:
```text
aws-lc-rs v1.18.1
aws-lc-fips-sys v0.14.2
```
(`aws-lc-rs` with `default-features = false, features = ["fips"]`.) This code emits both names. If config correctly attests `aws-lc-fips-sys`, `classifyRustCrypto` still rejects the `aws-lc-rs` wrapper as unattested. Attesting the wrapper separately would be wrong because it can select FIPS or non-FIPS providers.

Could the combined evidence map to one stable AWS-LC FIPS provider ID while preserving the `aws-lc-fips-sys` version?

internal/validations/validations.go
hexfusion: +1 updated the test to return ErrRustNoProvider.

internal/scan/node_scan.go
hexfusion: RPM path now runs validateModuleArtifactsPhase, in parity with walkDirScan.

internal/rust/rust.go
hexfusion: folded aws-lc-rs into aws-lc-fips-sys while keeping version preserved

smith-xyz:
@hexfusion thank you again for getting this in! been on my mind for awhile, we can adjust things as we go if needed.

smith-xyz (approved):

